PR #132, opened 2015-02-12
Use ansible from mirror and add retries to apt

git-harry:
There are a number of apt packages installed. These tasks can fail.

This commit modifies the ansible tasks to retry if there are failures
installing packages.


hughsaunders:
LGTM, but I'll wait for @jacobwagner or @jpmontez to have a look before merging.

stevelle:
nice patch. lgtm

hughsaunders:
@jacobwagner has volunteered to update this patch for current master :tada:

PR #135, opened 2015-02-18
Homogenize all the things

jacobwagner:
Massive amounts of work around getting the dev and qe plays / roles merged

- fixed numerous variable file issues
- refactored roles names to fit a a better syntax
- reworked all of the qe roles / plays to merge into some of the dev work


jacobwagner:
I probably need to squash this

roles/configure-rpc-compute/templates/target.sh.j2
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env bash
```
hughsaunders: may be able to replace this with https://github.com/stackforge/os-ansible-deployment/blob/master/scripts/run-playbooks.sh ?

hughsaunders:
Looks like a great start @jacobwagner, don't worry about breaking the commit builds at the moment, they've been broken since the os-ansible-deployment rename patch went in. I'll look into fixing that after Kevin's patch lands.

roles/configure-rpc-compute/templates/target.sh.j2
jacobwagner: Yeah I didnt even know all this existed but there is a lot of awesome functionality in those scripts that we could def utilize

odyssey4me:
Try to make use of the scripts as far as possible. You should find that they'll give you almost everything you need and, most importantly, they take away most of the differentials between branches/tags. If they don't I'd like to know about it in order to correct that.

dockerfiles/nightly-job/Dockerfile
```diff
@@ -10,7 +10,8 @@ RUN apt-get upgrade -y
 RUN apt-get install -y curl git python python-pip python-dev ssh-client
 
 # Install ansible
-RUN pip install ansible
+RUN git clone --recursive --branch v1.8.2 https://github.com/ansible/ansible
```
sigmavirus24: Probably this has been answered elsewhere, but why are we installing ansible from source?

odyssey4me: @cloudnull had a reason, although I can't remember what it was. Note that it would be better to make use of scripts/bootstrap-ansible.sh to prepare the environment appropriately for deployments (assuming you're able to in the jenkins scenario). If you do you'll end up using the same version of ansible, installed the same way, as would be done by a deployer. Each branch may support/require a different version of ansible.

cloudnull:
@sigmavirus24 in the past we were able to install ansible 1.6.x by simply pip installing ansible or pip installing a remote tarball, which is what we were doing.  Unfortunately this is no longer possible because upstream ansible has split the repository up into 3 main repos. With that change simply pip installing ansible left you with no core or extra modules. So it was generally useless at that point. As @odyssey4me mentions using the scripts takes care of all of that for the us and ensures that the installation process is the same for everyone.

sigmavirus24:
Makes sense. :+1: to using the bootstrap-ansible script.

roles/configure-rpc-compute/templates/target.sh.j2
cloudnull: +1 I think this script can be replaces with "https://github.com/stackforge/os-ansible-deployment/blob/master/scripts/run-playbooks.sh"

d34dh0r53: +1

jacobwagner:
closing this PR, see the qe_dev_converge branch

PR #149, opened 2015-03-06
Remove use of cidr.py

git-harry:
cidr.py is used to determine the IP addresses to be assigned to certain
vars. The tasks that use the script run it on the box running ansible.
The pip dependencies that are required by the script are being
installed on the wrong hosts. The complexity that this script brings is
not required by commit-multinode.yml because the IP addresses are
constructed based on a member_number var set in the inventory.

This commit removes cidr.py and the associated code and replaces it
with the use of the member_numbar var format already used by the
commit-multinode.yml vars.


hughsaunders:
LGTM @jacobwagner is this ok with you?

playbooks/roles/pip/tasks/main.yml
```diff
@@ -14,7 +14,7 @@
   retries: 3
 
 - name: Install pip
-  shell: "python /opt/get-pip.py --no-index --find-links={{ repo_url }}/python_packages/{{ targetBranch }}/"
+  shell: "python /opt/get-pip.py"
```
jacobwagner: i changed this cause it will install the latest version of pip which can break the versions of openstack we are installing. This mirrors how os-ansible-deploy installs pip to make sure that we are installing the openstack recommended version.

jacobwagner:
This change is ok. We will need to update the vars files for these labs and the range of used ips to match the member_number ips that we use

PR #156, opened 2015-03-09
The swift setup needs to run against all nodes

andymcc:
Swift drives were not being mounted on any but infra[0], this config
needs to happen on all nodes. Only the bootstrap should happen on
infra[0] and not the other nodes.

### DIFF
--- a/playbooks/commit-multinode.yml
+++ b/playbooks/commit-multinode.yml
@@ -33,17 +33,24 @@
       with_items: vgs
   tags: prepare
 
+- hosts: all
+  tags: prepare
+  user: root
+  roles:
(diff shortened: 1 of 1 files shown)


playbooks/commit-multinode.yml
```diff
@@ -33,17 +33,24 @@
...
   tags: prepare
 
-- hosts: infrastructure[0]
+- hosts: all
   tags: prepare
   user: root
   roles:
-
     - configure-rpc-swift-lvs
     - setup-git
```
jacobwagner: do you want to run setup git on all the hosts?

PR #170, opened 2015-03-12
Don't expect the pw-gen to create user_secrets

andymcc:
Since user_secrets already exists, expecting it to create the file will
mean it is skipped and the passwords are never populated.


andymcc:
Without this patch you get the following error:

ok: [node06] => {"changed": false, "cmd": "~/rpc_repo/scripts/pw-token-gen.py --file /etc/openstack_deploy/user_secrets.yml", "rc": 0, "stderr": false, "stdout": "skipped, since /etc/openstack_deploy/user_secrets.yml exists"}

PR #190, opened 2015-03-17
Revert "log results of os-ansible-script executions"

hughsaunders:
Reverts rcbops/jenkins-rpc#189

turns out that |& is a bashism and ansible is using another shell :(


Apsu:
You can set the executable to `/bin/bash` for the shell module if you like and can put this back in. See: http://docs.ansible.com/shell_module.html

PR #227, opened 2015-03-26
Jenkins Docker Wiring for HW Multinode

hughsaunders:
Adds Dockefile and job script for hardware multinode runs.

### DIFF
--- a/dockerfiles/commit-multinode/Dockerfile
+++ b/dockerfiles/commit-multinode/Dockerfile
@@ -1,0 +1,33 @@
+FROM ubuntu:14.04.2
+MAINTAINER Evan Callicoat <apsu@propter.net>
+
+# Make apt quiet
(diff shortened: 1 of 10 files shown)


Apsu:
Is there a reason why you're putting all kinds of python and ansible bits into a jenkins-slave container? You're putting all that into the multinode job container, and I assume you're not calling ansible directly from the jenkins job script (outside of a job container). Also, the Dockerfile 'style' is different between the two for how you're installing ansible, so that should be made consistent if there's some good reason you need it on both.

hughsaunders:
@apsu you make a good point. This PR actually contains bits of various attempts at the same problem. I'll tidy it up and remove the bits I'm not using.

hughsaunders:
In fact, I will create a new PR for the new version, so closing this one.

PR #238, opened 2015-04-01
Move multinode job logic into commit-multinode.sh

hughsaunders:
This patch simplifies the commit multinode gating strategy by moving
some of the logic that was previously in jenkins jobs into the
commit-multinode.sh script.

It also removes a few layers:
 * no longer spawns a docker instance for each job, all jobs run within
   the jenkins-slave container
 * osad playbooks are no longer called from a role within jenkin-rpc
   playbooks, they are called directly over ssh.

### DIFF
--- a/playbooks/roles/pip/tasks/main.yml
+++ b/playbooks/roles/pip/tasks/main.yml
@@ -4,14 +4,10 @@
     state: "directory"
     path: "/opt"
 
-- name: Get Modern PIP
-  get_url:
-    url: "{{ get_pip_url }}"
-    dest: "/opt/get-pip.py"
-    validate_certs: "no"
-  register: get_pip
-  until: get_pip|success
-  retries: 3
+- name: Get Modern Pip
+  shell: "for try in {1..10} :; do curl {{get_pip_url}} > /opt/get-pip.py; file /opt/get-pip.py | grep -q Python && break; sleep 30; done"
+  args:
+    executable: /bin/bash
 
 - name: Install pip
   shell: "python /opt/get-pip.py"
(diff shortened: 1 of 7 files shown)


hughsaunders:
tested: http://jenkins.propter.net/job/commit-multinode-script-logic/206/

Apsu:
Can one of the admins verify this patch?

hughsaunders:
ignore the commit status, it was an attempt at setting up a gate for this repo, but wasn't successfull.

playbooks/roles/pip/tasks/main.yml
```diff
@@ -4,14 +4,10 @@
...
-- name: Get Modern PIP
-  get_url:
-    url: "{{ get_pip_url }}"
-    dest: "/opt/get-pip.py"
-    validate_certs: "no"
-  register: get_pip
-  until: get_pip|success
-  retries: 3
+- name: Get Modern Pip
+  shell: "for try in {1..10} :; do curl {{get_pip_url}} > /opt/get-pip.py; file /opt/get-pip.py | grep -q Python && break; sleep 30; done"
```
jacobwagner: i get what this is doing but why not just up the retry count from 3 to 10?

jacobwagner:
besides the above question. looks good to me +1

playbooks/roles/pip/tasks/main.yml
hughsaunders: @jacobwagner the server sometimes serves empty responses with an OK response code :(
This line will retry if the "file" command doesn't detect that the response is a python script.

jacobwagner: sounds good, +1 to the merge

PR #271, opened 2015-04-17
Juno upgrade

hughsaunders:


playbooks/roles/pip/tasks/main.yml
```diff
@@ -4,17 +4,13 @@
...
-  until: get_pip|success
-  retries: 3
+- name: Get Modern Pip
+  shell: "for try in {1..10} :; do curl {{get_pip_url}} > /opt/get-pip.py; file /opt/get-pip.py | grep -q Python && break; sleep 30; done"
+  args:
+    executable: /bin/bash
 
 - name: Install pip
-  shell: "python /opt/get-pip.py --no-index --find-links={{ repo_url }}/python_packages/{{ os_ansible_branch }}/"
+  shell: "python /opt/get-pip.py"
```
jacobwagner: This breaks things for QE. We get versions of pip that dont work with icehouse. Since we are still testing icehouse this will cause massive breaks in those tests

PR #307, opened 2015-05-07
Adjust tags to run all required plays in one.

andymcc:
This change ensures that a failure on a specific play will cause the
step to fail instead of succeeding because, for example, the reboot play
(which is last) succeeds despite the other plays all failing.

### DIFF
--- a/playbooks/nightly-multinode.yml
+++ b/playbooks/nightly-multinode.yml
@@ -38,15 +38,26 @@
     - setup-git
 
 - hosts: infrastructure[0]
+  tags: prepare
   user: root
-  tags: configure
+  roles:
+    - role: run-script-from-os-ansible-deployment
+      script_name: bootstrap-ansible
+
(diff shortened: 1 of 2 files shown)


playbooks/nightly-multinode.yml
```diff
@@ -38,15 +38,26 @@
     - setup-git
 
 - hosts: infrastructure[0]
+  tags: prepare
   user: root
-  tags: configure
+  roles:
+    - role: run-script-from-os-ansible-deployment
+      script_name: bootstrap-ansible
+
```
jacobwagner: i removed this and added it as a straight script call in the nightly bash script due to the output and redirect in jenkins. When you have jenkins calling bash, which then calls ansible, which in turn runs more bash..bad things happen

jacobwagner:
Merging this. I will deal with the redirect issue some other way as failing on an actual failure is better than  not failing

PR #317, opened 2015-05-08
Disable tempest_compute_run_ssh

mattt416:
Currently, a number of tempest compute api tests are failing as
instances are not being brought up correctly with a floating IP
address.  This makes ssh'ing into the instances impossible and therefore
we need to set tempest's run_ssh option to False.

Once tempest is updated to properly create floating IP addresses for
all tests then we should be able to flip this variable back to True.


andymcc:
LGTM

jacobwagner:
+1

PR #324, opened 2015-05-11
Add Finally Tags

hughsaunders:
Finally tags is a new parameter that specifies a list of tags that run
after the initial tags list ($TAGS) reagrdless of the exit status of the
first list.

This enables the job to break out of the $TAGS list after any failure
but still continue on to $FINALLY_TAGS and do any info collection /
cleanup tasks that are required.


jacobwagner:
+1

PR #358, opened 2015-06-04
Auto verify github ssh key

hughsaunders:
* Remove strict host key checking
* Enable DNS Host key verification
* set accept_hostkey when using git module


odyssey4me:
LGTM, and verified to work using @hughsaunders' fork

PR #363, opened 2015-06-05
Dont accumulate return codes

hughsaunders:
Currently the final return code is the sum of previous return codes.
This can be problematic as if the return codes sum to 256, that will be
intepreted as success when it actually represents multiple failures.

This patch ensures the return code is 0 for success or 1 for failure
regardless of the number of failures.


hughsaunders:
check this run before merging: http://jenkins.propter.net/job/commit-multinode/996/console

jacobwagner:
lgtm +1

PR #378, opened 2015-06-08
Remove default for TAGS, update comments

hughsaunders:
The TAGS default was incorrect (CSV, should have been space separated).
Better to remove the default for tags so the user is not surprised when
running a job with only FINALLY_TAGS.

Comments are updated, and write_properties is no longer called before
tags, as claim is now a tag, so no further information can be known
before at least one tag as run.


hughsaunders:
Test run: ensure this passes before merging: http://jenkins.propter.net/job/commit-multinode/1025/console

PR #383, opened 2015-06-11
[Juno] Improve container destruction

jpmontez:
Force deletion of all LXC containers regardless of their state. This makes error-handling in environment cleanup much more efficient when containers are stuck building.


jpmontez:
http://jenkins.propter.net/job/release-multinode-docker/217/console

PR #386, opened 2015-06-12
Install heatclient in jenkins master dockerfile

hughsaunders:
Heatclient is required for running the heat-multinode jobs.


jacobwagner:
lgtm +1

PR #389, opened 2015-06-22
Refactor

jpmontez:
- Remove RPC references within roles, variables, playbooks
- Consolidate multinode scripts, playbooks, and Dockerfiles

Add a common Dockerfile for all multinode environments. Further distinctions are made in the `scripts/multinode.sh` script in separating types of labs and how roles are applied within Ansible.

Aiding in this, prefixes were added to customer-inspired environment inventory and variable files. Variable names within bash scripts were also fleshed out to help with self-documentation.


scripts/multinode.sh
```diff
@@ -0,0 +1,153 @@
...
+}
+
+function cleanup {
+  run_tag cleanup
+}
+
+function main {
+  local retval
+
+  # Variables based products' respective documentation
```
jacobwagner: we do different type of string comparisons here...why? Was this a typo on my part?

jpmontez: What do you mean by different types, exactly? I changed the product names to be more specific. Also, I included an `elif` because I thought it should fail hard if the name was improperly set rather than fail later on down the road.

jacobwagner: i meant we do = comparison and == comparisons. while tech the same thing it could get confusing / redundant

jpmontez: I changed the other string comparison.

jacobwagner:
This refactor does not work with the nightly labs, closing it (will reopen with fixed PR)

PR #416, opened 2015-07-02
[kilo] Remove 'openstack_repo_url' reference

jpmontez:
Pip repositories are now built within OSAD. Therefore, we don't need this variable anymore.

### DIFF
--- a/playbooks/vars/nightly-juno.yml
+++ b/playbooks/vars/nightly-juno.yml
@@ -64,7 +64,7 @@ swift_config:
   weight: 100
   min_part_hours: 1
   repl_number: 3
-  storage_network: 'br-storage' 
+  storage_network: 'br-storage'
   replication_network: 'br-repl'
   drives:
     - name: swift1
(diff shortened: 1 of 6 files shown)


playbooks/roles/configure-compute/templates/user_variables.j2
```diff
@@ -1,7 +1,4 @@
 glance_default_store: swift
-{% if repo_url is defined %}
-openstack_repo_url: '{{ repo_url }}'
-{% endif %}
```
jacobwagner: done remove this, it is a conditional that needs to stay in place in case we ever want to go back to using a repo_url

playbooks/vars/nightly-icehouse.yml
```diff
@@ -1,6 +1,5 @@
 config_prefix: rpc
 repo_dir: rpc_repo
-repo_url: https://rpc-repo.rackspace.com
```
jacobwagner: leave this as icehouse will always use it

PR #417, opened 2015-07-07
Implement native modules

jpmontez:
Should solve issue: https://github.com/rcbops/jenkins-rpc/issues/401

Set role syntax to match OSAD's style guide. Use native git and LXC Ansible modules.


jpmontez:
Depends on PR https://github.com/rcbops/jenkins-rpc/pull/421

PR #418, opened 2015-07-08
deploy tempest by default on all nightly and release labs

jacobwagner:


jpmontez:
This variable would only work with OSAD deployments.

`rpc-openstack` doesn't recognize this environment variable. The playbook would have to be run manually or included in the `deploy.sh` script.

PR #421, opened 2015-07-08
Change MaaS default

jpmontez:
This allows the `DEPLOY_MAAS` variable to be overridden within a Jenkins job. This needs to be present for `rpc-openstack` builds to complete successfully within IAD environments.


jacobwagner:
This will set deploy maas to yes by default and will also set it to deploy in the nightly labs which will break the nightly labs...

jpmontez:
Fixed.

PR #426, opened 2015-07-08
[Kilo] Add Travis-CI Integration

jpmontez:
Once turned on in the Travis-CI dashboard, we'll have basic linting of our Ansible playbooks.


nrb:
It appears I've duplicated some of the linting work in https://github.com/rcbops/rpc-openstack/issues/240.

We should probably collaborate on this.

PR #471, opened 2015-11-18
Ceph integration

jacobwagner:
- added ability to deploy ceph
- added storage01 lab
- added ability to split out rpc features
- added more tags to allow greater control of prepare configuration of labs


jpmontez:
Looks good except for the addition of `RPC_FEATURES`. I don't like that `qe-iad3-lab01` is hard-coded with features as a result of how that variable is implemented.

We already know that Ceph should be configured because it's defined within the laboratory's variable file. Can you adjust the playbook to conditionally apply the role based on whether `ceph` is defined in the inventory or variable file?

- http://docs.ansible.com/ansible/playbooks_roles.html#roles
- http://docs.ansible.com/ansible/playbooks_conditionals.html#the-when-statement

PR #492, opened 2016-01-21
made tags a list and convert to string before passing

jacobwagner:


scripts/multinode.sh
```diff
@@ -108,6 +110,27 @@ function run_tag {
   fi
 }
 
+function build_tags {
+  # This is a nasty method that could be done much better: jwagner
+  # We need to create comma seperated no space string of all tags
+  # that we will run, ignoreing empty tag values
+  TAGS=$tag
+
+  echo $TAGS
```
jacobwagner: ill remove this later

PR #567, opened 2016-04-11
Add script to calculate upgrade paths

git-harry:
rpc-openstack supports two types of upgrade, major upgrades using
upgrade.sh and minor/patch upgrades where running a newer version of the
playbooks should result in the same/compatible configuration deployed.

This commit adds a script to determine the supported upgrade paths for
any commit. It assumes that x.0.z does not support major upgrades and
that minor version can only upgrade between contiguous releases.

master is calculated to be the next major release if the
closest openstack-ansible tag to the current SHA is on the next major
release to rpc-o. If it is not the next major it is assumed to be the
next minor release.

### DIFF
--- a/scripts/upgrade-branches.sh
+++ b/scripts/upgrade-branches.sh
@@ -1,0 +1,103 @@
+#!/bin/bash
+
+#Assume in rpc-openstack directory and checked out on commit to test rebased on appropriate branch
+
+function check_tag {
+  # Tag sorting makes rc versions newer than the released version
+  # This function returns the release if it exists else the rc
+  tag_to_test=$1
+  t=''
+  if [[ $(echo $tag_to_test | egrep '^r[0-9]+\.[0-9]+\.[0-9]+rc[0-9]+$') ]]; then
+    t=$(git tag --list $(echo $1 | egrep -o '^r[0-9]+\.[0-9]+\.[0-9]+'))
+  fi
+  if [[ $t != '' ]]; then
+    echo $t
+  else
+    echo $tag_to_test
+  fi
+}
+current_branch=$1
+
+closest_tag=$(git describe --tags --abbrev=0 origin/$current_branch)
+
+original_ref=$(git rev-parse 'HEAD^{commit}')
+
+git checkout --quiet master
+git submodule --quiet update --init
+pushd openstack-ansible > /dev/null
+  osa_t=$(check_tag $(git describe --abbrev=0))
+popd > /dev/null
+git checkout --quiet $original_ref
+git submodule --quiet update --init
+
+osa_major_number=$(echo "$osa_t" | cut -d. -f1)
+# Assume rpc-openstack major version is the same as OpenStack-Ansible tag major version.
+master_major_number=$osa_major_number
+
+if [[ $current_branch == 'master' ]]; then
+  # If the branch/rc tag process was not followed correctly closest_tag might not show what was expected
+  #c_tag=($(echo $closest_tag | egrep -o '[0-9]+\.[0-9]+\.[0-9]+' | cut -d. --output-delimiter=' ' -f1-))
+  c_tag=($(git tag --list r[0-9]*| sort -V | tail -n1 | egrep -o '[0-9]+\.[0-9]+\.[0-9]+' | tr '.' ' '))
+  branch_major_number=${c_tag[0]}
+  if [[ ${c_tag[0]} == $master_major_number ]]; then
+    # master should be for the next minor after the one pointed to be closest_tag
+    branch_minor_number=$(expr ${c_tag[1]} + 1)
+  else
+    # master should be for a new major
+    branch_major_number=$master_major_number
+    branch_minor_number=0
+  fi
+  current_patch_version=''
+else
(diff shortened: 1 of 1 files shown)


scripts/upgrade-branches.sh
```diff
@@ -0,0 +1,102 @@
+#!/bin/bash
+
+#Assume in rpc-openstack directory and checked out on commit to test rebased on appropriate branch
+
+function check_tag {
+  # Tag sorting makes rc versions newer than the released version
+  # This function returns the release if it exists else the rc
+  t=''
+  if [[ $(echo $1 | egrep '^r[0-9]+\.[0-9]+\.[0-9]+rc[0-9]+$') ]]; then
```
hughsaunders: if egrep '^r[0-9]+\.[0-9]+\.[0-9]+rc[0-9]+$' <<<$1; then

scripts/upgrade-branches.sh
```diff
@@ -0,0 +1,102 @@
+#!/bin/bash
+
+#Assume in rpc-openstack directory and checked out on commit to test rebased on appropriate branch
+
+function check_tag {
```
hughsaunders: please define function params by giving them names eg commitish="$1"

scripts/upgrade-branches.sh
```diff
@@ -0,0 +1,102 @@
...
+git submodule update --init
+
+osa_major_number=$(echo "$osa_t" | cut -d. -f1)
+# Assume rpc-openstack major version is the same as OpenStack-Ansible tag major version.
+master_major_number=$osa_major_number
+
+if [[ $current_branch == 'master' ]]; then
+  # If the branch/rc tag process was not followed correctly closest_tag might not show what was expected
+  #c_tag=($(echo $closest_tag | egrep -o '[0-9]+\.[0-9]+\.[0-9]+' | cut -d. --output-delimiter=' ' -f1-))
+  c_tag=($(git tag --list r[0-9]*| sort -V | tail -n1 | egrep -o '[0-9]+\.[0-9]+\.[0-9]+' | cut -d. --output-delimiter=' ' -f1-))
```
hughsaunders: `| tr '.' ' '` may be slightly more obvious than cut

scripts/upgrade-branches.sh
```diff
@@ -0,0 +1,102 @@
...
+  branch_major_number=${c_tag[0]}
+  if [[ ${c_tag[0]} == $master_major_number ]]; then
+    # master should be for the next minor after the one pointed to be closest_tag
+    branch_minor_number=$(expr ${c_tag[1]} + 1)
```
hughsaunders: branch_minor_number=$(( ++c_tag[1] ))

git-harry: I don't want to change the values in c_tag in case the script gets changed in the future and wants to make further use of that variable.

scripts/upgrade-branches.sh
```diff
@@ -0,0 +1,102 @@
...
+  fi
+  if [[ $t != '' ]]; then
+    echo $t
+  else
+    echo $1
+  fi
+}
+current_branch=$1
+
+closest_tag=$(git describe --tags --abbrev=0 origin/$current_branch)
```
hughsaunders: Should probably fetch origin before this

scripts/upgrade-branches.sh
git-harry: Done

scripts/upgrade-branches.sh
git-harry: Done

scripts/upgrade-branches.sh
```diff
@@ -0,0 +1,102 @@
...
+  osa_t=$(check_tag $(git describe --abbrev=0))
+popd
+git checkout $original_ref
+git submodule update --init
+
+osa_major_number=$(echo "$osa_t" | cut -d. -f1)
+# Assume rpc-openstack major version is the same as OpenStack-Ansible tag major version.
+master_major_number=$osa_major_number
+
+if [[ $current_branch == 'master' ]]; then
```
hughsaunders: Comment here to explain that if current branch is master, then tags must be used to determine the major, minor and patch version of the current branch, otherwise the major and minor version can be determined from the branch name.

scripts/upgrade-branches.sh
```diff
@@ -0,0 +1,102 @@
...
+fi
+next_major_version=$(git tag --list r$(expr ${branch_major_number} + 1).* | sort -V | tail -n1)
+if [[ $current_branch != 'master' ]] && [[ $next_major_version == '' ]] && [[ $master_major_number -gt $branch_major_number ]]; then
+  next_major_version=master
+fi
+
+previous_minor_version=$(check_tag $(git tag --list r${branch_major_number}.$(expr ${branch_minor_number} - 1).* | sort -V | tail -n1))
+previous_major_version=$(git tag --list r$(expr ${branch_major_number} - 1).* | sort -V | tail -n1)
+
+#echo "Branch: $current_branch"
```
hughsaunders: This section could be useful, how about outputting it to stderr?

git-harry: Done.

PR #570, opened 2016-04-13
Find all paths when no branch specified

hughsaunders:


sigmavirus24:
Looks like a good minor refactor to pull some code into a reusable function. :+1:

scripts/upgrade-branches.sh
```diff
@@ -16,88 +16,103 @@ function check_tag {
...
+  elif [[ $next_minor_version != '' ]]; then
+    test_branches="$test_branches minor-to-$next_minor_version"
+  fi
+
+  echo $test_branches
+}
+
+# --- Main ---
+branch="$1"
+if [[ -z "$branch" ]]; then
```
git-harry: I think the output from the script should always be in the same format. I'd suggest moving the for-loop out of the conditional and have branches be a list of one branch if branch is already defined.

hughsaunders: Done

scripts/upgrade-branches.sh
```diff
@@ -16,88 +16,103 @@ function check_tag {
...
+  branches=$(git branch -a |awk -F'[/ ]' '/[a-z]+-[0-9]+\.[0-9]+$/{print $NF}' |sort -u)
```
git-harry: This doesn't include master

hughsaunders: Done

scripts/upgrade-branches.sh
```diff
@@ -16,88 +16,103 @@ function check_tag {
...
+    test_branches="$test_branches minor-to-$next_minor_version"
+  fi
+
+  echo $test_branches
+}
+
+# --- Main ---
+branch="$1"
+if [[ -z "$branch" ]]; then
+  branches="master $(git branch -a |awk -F'[/ ]' '/[a-z]+-[0-9]+\.[0-9]+$/{print $NF}' |sort -u)"
```
git-harry: I fetched all the remote branches and now the scripts is outputting bug branches, the following should fix it.
```
git branch -a |awk -F'[/ ]' '/\/[a-z]+-[0-9]+\.[0-9]+$/{print $NF}' |sort -u
```

hughsaunders: Good spot, I hadn't fetched all branches in my test repo. Done

sigmavirus24:
:+1:

PR #574, opened 2016-04-15
Build Summary: More failure matchers

hughsaunders:


scripts/build-summary/build.py
```diff
@@ -132,6 +135,30 @@ def open_log(filename):
         if not self.failures:
             self.failures.add("Unknown Failure")
 
+    def tempest_filter_fail(self, lines):
+        match_re = re.compile("'Filter (.*) failed\.")
+        for i, line in enumerate(lines):
```
git-harry: i doesn't seem to be used in any of these functions so I think you can get rid of enumerate

hughsaunders: "i" is used to pass the failure line to get_previous_task, however in 3/4 of the new additions get_previous task isn't used so I'll remove enumerate as suggested.

scripts/build-summary/build.py
```diff
@@ -132,6 +135,30 @@ def open_log(filename):
...
+            match = match_re.search(line)
+            if match:
+                self.failures.add("Openstack Tempest Gate test "
+                                  "set filter {fail} failed.".format(
+                                      fail=match.groups()[0]))
```
git-harry: I believe 'groups()[0]' is equivalent to 'group(1)'.

hughsaunders: Done

PR #576, opened 2016-04-18
Add upgrade parent job script

hughsaunders:
Script added here so that as little as possible is in the actual jenkins
config. This script takes the output of upgrade branches, converts
references to absolute and writes a properties file which will be fed to
the matrix job for generating the dynamic axis.

Also changed the delimiter used by upgrade-branches from - to _ as some
branch names contain -.

### DIFF
--- a/scripts/dev-aio-upgrade/upgrade-parent-job.sh
+++ b/scripts/dev-aio-upgrade/upgrade-parent-job.sh
@@ -1,0 +1,9 @@
+#!/bin/bash -x
+
+rm -rf rpc-openstack
+git clone --recursive $RPC_REPO
+
+pushd rpc-openstack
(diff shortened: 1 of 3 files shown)


scripts/dev-aio-upgrade/upgrade-parent-job.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash -x
+
+rm -rf rpc-openstack
```
git-harry: Does $RPC_REPO == 'rpc-openstack'?

git-harry: Wait, duh! Ignore me.

PR #582, opened 2016-04-27
Sync submodules

git-harry:
If a submodule URL is updated in .gitmodules by a patch being tested or
between versions being used for an upgrade test the submodule's URL in
.git/config is not updated.

This commit adds 'git submodule sync' to the build script so that
.git/config is correct prior to updating the submodules.

### DIFF
--- a/scripts/aio_build_script.sh
+++ b/scripts/aio_build_script.sh
@@ -40,6 +40,7 @@ else
     }
 fi
 
+git submodule sync
 git submodule update --init
 
 # git plugin checks out repo to root of workspace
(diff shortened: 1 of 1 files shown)


mattt416:
👍

scripts/aio_build_script.sh
```diff
@@ -40,6 +40,7 @@ else
     }
 fi
 
+git submodule sync
 git submodule update --init
```
evrardjp: won't git submodules sync skip the sync for the submodules that aren't initialized yet?
my point is maybe the line should be split it:

git submodule init
git submodule sync
git submodule update

I am not a submodules expert, so I trust you on this.

evrardjp: I think in all cases it should work if it wasn't initialized, so I think it could be merged.

git-harry: If the submodule hasn't been initialised there is no need to sync because there is no .git/config entry to update.

Jenkins has already initialised the submodule prior to this script being executed and so this sync is needed to account for that.

PR #593, opened 2016-06-08
Add holland check

hughsaunders:
Also adds ability to override the Openstack Ansible submodule. 

Holland check verified by manually started job: http://jenkins.propter.net/job/RPC-AIO/7058/console

Regression test for rcbops/rpc-openstack#1121

### DIFF
--- a/scripts/aio_build_script.sh
+++ b/scripts/aio_build_script.sh
@@ -39,6 +39,16 @@ run_holland(){
   sudo lxc-attach -n $(sudo lxc-ls |grep galera|head -n1) -- /bin/bash -c "holland bk"
 }
 
+override_oa(){
+  if [[ "$OA_REPO" != "none" ]]; then
+    pushd openstack-ansible
+      git remote add override $OA_REPO -f
+      git fetch -a
+      git checkout override/$OA_BRANCH
+    popd
+  fi
+}
+
 #fix sudoers because jenkins jcloud plugin stamps on it.
 sudo tee -a /etc/sudoers <<ESUDOERS
 %admin ALL=(ALL) ALL
@@ -79,6 +89,9 @@ fi
 
 git submodule sync
 git submodule update --init
+
+override_oa
+
 log_git_status
 
 # git plugin checks out repo to root of workspace
@@ -153,6 +166,7 @@ if [ "$UPGRADE" == "yes" ] && [ "$OVERALL_RESULT" -eq 0 ];
     fi
     git submodule sync
     git submodule update --init
+    override_oa
     log_git_status
     echo "********************** Run RPC Deploy/Upgrade Script ***********************"
     if [[ "$UPGRADE_TYPE" == "major" ]]; then
--- a/scripts/aio_build_script.sh
+++ b/scripts/aio_build_script.sh
@@ -6,6 +6,27 @@ log_git_status(){
   git branch -v
 }
 
+run_rpc_deploy(){
+  sudo \
+    DEPLOY_AIO=yes \
+    DEPLOY_HAPROXY=yes \
+    DEPLOY_TEMPEST=yes \
+    DEPLOY_CEPH=${DEPLOY_CEPH} \
+    DEPLOY_SWIFT=${DEPLOY_SWIFT} \
+    DEPLOY_MAAS=${DEPLOY_MAAS} \
+    ANSIBLE_GIT_RELEASE=ssh_retry \
+    ANSIBLE_GIT_REPO="https://github.com/hughsaunders/ansible" \
+    ADD_NEUTRON_AGENT_CHECKSUM_RULE=yes \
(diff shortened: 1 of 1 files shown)


git-harry:
Separate pull requests for the commits would be helpful

git-harry:
Ah I see, this is mostly a refactor with a holland thing thrown it. It would have been much easier if this had been separated in a holland and a refactor patch.

scripts/aio_build_script.sh
```diff
@@ -93,38 +139,24 @@ export BOOTSTRAP_OPTS="${BOOTSTRAP_OPTS} bootstrap_host_ubuntu_security_repo=${U
...
+run_tempest
 TEMPEST_RC=$?
 
-[ $DEPLOY_RC == 0 -a $TEMPEST_RC == 0 ]
+echo "********************** Run Holland ***********************"
+run_holland
+HOLLAND_RC=$?
+
+[ $DEPLOY_RC == 0 -a $TEMPEST_RC == 0 -a $HOLLAND_RC == 0 ]
 OVERALL_RESULT=$?
```
git-harry: I think the above two lines are potentially going to be missed now there is a calc_result function. Maybe you should split calc_result and have two function - calc_result and print_result. calc_result would be used to calculate OVERALL_RESULT and print_result would be used to print the all the return codes and the overall result. That would then allow you to replace the above with calc_result and also use it in print_result.

scripts/aio_build_script.sh
```diff
@@ -93,38 +139,24 @@ export BOOTSTRAP_OPTS="${BOOTSTRAP_OPTS} bootstrap_host_ubuntu_security_repo=${U
...
-  DEPLOY_CEPH=${DEPLOY_CEPH} \
-  DEPLOY_SWIFT=${DEPLOY_SWIFT} \
-  DEPLOY_MAAS=${DEPLOY_MAAS} \
-  ANSIBLE_GIT_RELEASE=ssh_retry \
-  ANSIBLE_GIT_REPO="https://github.com/hughsaunders/ansible" \
-  ADD_NEUTRON_AGENT_CHECKSUM_RULE=yes \
-  BOOTSTRAP_OPTS=$BOOTSTRAP_OPTS \
-  scripts/deploy.sh
-
+run_rpc_deploy
```
git-harry: What about having the functions return $? instead of having a separate line?

hughsaunders: I'm not sure I understand.. functions implicitly return $?

git-harry: Sorry, bash is hard. I guess I just thought something along the lines of

DEPLOY_RC=$(run_rpc_deploy)

would look nicer, assuming that will execute the function.

hughsaunders: I guess the function could output $?, then it could be captured with $(), but that doesn't really seem any better. The function could set FOO_RC itself.

git-harry: Yeah, maybe that would be better. I don't think it's a blocker the way it currently is, it just looks a bit strange to my uneducated eyes.

mancdaz: seems equally as kludgy as just doing it over 2 lines if you ask me

mancdaz:
works for me once the merge conflict is resolved @hughsaunders

hughsaunders:
reshuffled commits, test job: http://jenkins.propter.net/job/RPC-AIO/7294/

mancdaz:
👍

scripts/aio_build_script.sh
```diff
@@ -93,39 +139,23 @@ export BOOTSTRAP_OPTS="${BOOTSTRAP_OPTS} bootstrap_host_ubuntu_security_repo=${U
...
 TEMPEST_RC=$?
 
-[ $DEPLOY_RC == 0 -a $TEMPEST_RC == 0 ]
+echo "********************** Run Holland ***********************"
+run_holland
+HOLLAND_RC=$?
+
+echo "********************** Deployment Result **********************"
+calc_result
 OVERALL_RESULT=$?
```
git-harry: calc_result already sets OVERALL_RESULT, I'm not sure this line will do the right thing even if it was required.

scripts/aio_build_script.sh
```diff
@@ -6,6 +6,49 @@ log_git_status(){
...
+run_holland(){
+  sudo lxc-attach -n $(sudo lxc-ls |grep galera|head -n1) -- /bin/bash -c "holland bk"
+}
+
+override_oa(){
+  if [[ "$OA_REPO" != "none" ]]; then
+    pushd openstack-ansible
+      git remote add override $OA_REPO -f
+      git fetch -a
+      git checkout $OA_BRANCH
```
git-harry: I wonder if that should be:

```
git checkout override/$OA_REPO
```

git-harry:
👍  on the code, waiting on Hugh to confirm testing was successful.

hughsaunders:
👍

PR #597, opened 2016-06-22
set python unbuffered

hughsaunders:


git-harry:
@hughsaunders love the commit message

PR #599, opened 2016-06-22
Add Preserve Environment flag to sudo.

hughsaunders:
The rpc deploy script is called via sudo, which by default resets the
environment. This means that PYTHONUNBUFFERED=1 does not influce the
deploy script as intended.

This commit adds -E (preserve environment) to the relevant sudo
invocations.


evrardjp:
👍  Good to me

PR #601, opened 2016-06-23
Report top level job instead of immediate parent

hughsaunders:
Currently there is a three level chain for aio gate jobs:
*-PR --> Docs-Test --> RPC-AIO.

The build summary page reports the Docs-Test job as the parent for AIO
jobs, this is not very helpful as it doesn't communicate whether the job
is an upgrade, or ceph/swift etc. Reporting the top level job ensures
this information is reported however the hierarchy is modified.


d34dh0r53:
:+1: * 10000

PR #602, opened 2016-06-24
Show entire job hierachy in build summary.

hughsaunders:
There are situations in which the whole hierachy is useful, for example
in the PR jobs, the top level job is useful to show which type of job is
running, whereas in matrix builds this information is in the name of an
itermediate job.

This commit removes the job/parent job fields and replaces them with a
list of jobs starting from the root trigger.

Screenshot of new layout:

screen shot 2016-06-24 at 14 52 20

### DIFF
--- a/scripts/build-summary/build.py
+++ b/scripts/build-summary/build.py
@@ -55,22 +55,47 @@ class Build(object):
         return kvs
 
     def get_parent_info(self):
-        try:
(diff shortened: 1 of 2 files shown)


evrardjp:
Can't vote on the code, but I love it already.

alextricity25:
:+1:

PR #604, opened 2016-06-29
Make sure variables set by jenkins get priority

prometheanfire:


scripts/aio_build_script.sh
```diff
@@ -103,7 +103,7 @@ sudo ln -s $PWD /opt/rpc-openstack
 
 
 ## Add MAAS credentials
-uev=/opt/rpc-openstack/rpcd/etc/openstack_deploy/user_extras_variables.yml
+uev=/opt/rpc-openstack/rpcd/etc/openstack_deploy/user_zzz_extras_variables.yml
```
git-harry: Why not call it user_rpco_variables_overrides?

prometheanfire: We could, I chose zzz to always be the top priority, which I think is more important.

prometheanfire:
Renamed the file, hopefully this lets @git-harry be happy.  Also made sure the vars file had the proper header `---`.

sigmavirus24:
Commit message:

> priprity

Wat

prometheanfire:
fixed

sigmavirus24:
Looks fine to me.

git-harry:
The commit message is rather terse however the change should work as expected.

PR #605, opened 2016-06-30
Update build summary tooling to display liberty jobs

sigmavirus24:
This updates both the build library and the buildsummary template to
include the periodic jobs for our liberty branch.

Connects #156 

---

You can view an example of this at: http://jenkins.propter.net:443/test.html


scripts/build-summary/buildsummary.j2
```diff
@@ -123,18 +123,20 @@ $(document).ready( function () {
   <h3> Periodic Build Success (Last 48 Hours)</h3>
   <table id="summary" class="table">
     <thead>
-      <tr><th></th><th>Master</th><th>Kilo</th></tr>
+      <tr><th></th><th>Master</th><th>Kilo</th><th>Liberty</th></tr>
```
git-harry: It might be nicer to have them in chronological order i.e. Master, Liberty, Kilo.

git-harry:
In principle the code looks fine, I'm not completely familiar with the working of these scripts to rely on my eyes as the sole test. Looking at the test page you've linked to, the other sections aren't functioning correctly, they don't have searchable tables like the real page. Is this to be expected?

sigmavirus24:
> Looking at the test page you've linked to, the other sections aren't functioning correctly, they don't have searchable tables like the real page. Is this to be expected?

I didn't test that, but I also didn't change anything that should affect that. I expect once it's deployed as it's expected it will continue to work.

scripts/build-summary/buildsummary.j2
sigmavirus24: Makes sense. Will update with this.

sigmavirus24: Fixed

sigmavirus24:
Maybe @Apsu can review this for correctness?

Apsu:
@sigmavirus24 This is all new to me since last I touched Jenkins (18+ months ago?), so... don't know for sure what it's supposed to do/look like/have. However, your commits appear innocuous, assuming the _/. replacements and extra column aren't affecting some other fragile piece and weirdly breaking it, which seems unlikely.

sigmavirus24:
The underscore work is purely because I don't trust Jinja2 to do the right thing. Call it paranoia. :) (Also call it hatred of the current branch naming scheme of `{codename}-{major_number}.{minor_number}`.)

PR #606, opened 2016-07-01
Remove use of sudo when updating variables file

git-harry:
The commit bfd4dbb5bc0df7d5a0b14753637d71dffabeee52 introduced the use
of a new file, user_zzz_gating_variables.yml, as the place for Jenkins
to set any job-specific Ansible variables. This file, unlike the
previously used one, user_extras_variables.yml, does not exist by
default. When aio_build_script.sh attempted to modify it, this caused it
to be created with sudo, causing it to be owned by
root, which meant that a later command was failing when trying to modify
the file without using sudo. The result of this is that the build does
not have all the expected variables set.

This commit removes the use of sudo so that the file is owned by the
user jenkins and group wheel.


prometheanfire:
:+1: makes sense

sigmavirus24:
👍

PR #608, opened 2016-07-04
Show multinode results as master.

hughsaunders:
While the template in use is rpc-12, the rpc sha is overridden to
master, so the result should be shown in the master column.


hughsaunders:
recheck_all

hughsaunders:
recheck_all

PR #609, opened 2016-07-06
First attempt at JJB definitions

hughsaunders:
Jenkins job builder constructs/updates jenkins job from yaml defintions.

This PR includes three jobs:
* JJB-Jenkins-RPC-PR-aio & JJB-Jenkins-RPC-PR-upgrade
    These jobs test changes to jenkins-rpc by running RPC-AIO as a standard
    aio and as an upgrade.
* JJB-Job-Update
    This job runs jjb to ensure that the jobs in jenkins match the yaml
    job defintions. This job is triggered post merge instead of on pr,
    so that live jobs aren't updated before they are approved.

Before this PR is merged, JJB must be run manually on the jenkins master
as the jobs dir doesn't yet exist in jenkins-rpc/master so
JJB-Job-Update will fail.
commit fa2608845a188c76b051114b0c10adb3025a56cc

Connects https://github.com/rcbops/u-suk-dev/issues/222

### DIFF
--- a/rpc-jobs/jenkins_jobs.ini
+++ b/rpc-jobs/jenkins_jobs.ini
@@ -1,0 +1,12 @@
+[job_builder]
+ignore_cache=True
+keep_descriptions=False
+include_path=.
+recursive=False
+exclude=.*:manual:./development
+allow_duplicates=False
+
+[jenkins]
+user=rcb-ro
+url=http://jenkins.propter.net
+query_plugins_info=False
--- a/rpc-jobs/jobs.yaml
+++ b/rpc-jobs/jobs.yaml
@@ -1,0 +1,97 @@
+## Macros for repeated blocks
+
+# Define the github project associated with jenkins rpc
+- property:
+    name: jenkins-rpc-github
+    properties:
+      - github:
+          url: https://github.com/rcbops/jenkins-rpc
+
+# Define the scm/git repo associated with jenkins-rpc
+- scm:
+    name: jenkins-rpc-git
+    scm:
+      - git:
+          url: https://github.com/rcbops/jenkins-rpc
+          branches:
+            - master
+          refspec: "+refs/pull/*:refs/remotes/origin/pr/* +refs/heads/*:refs/remotes/origin/*"
+          name: origin
+
+# This project instantiates the JJB-Jenkins-RPC-PR-{type} template twice
+# to create an aio and an upgrade job.
+- project:
+    name: JJB-Jenkins-RPC-PR-Jobs
+    jobs:
+      - 'JJB-Jenkins-RPC-PR-{type}':
+          type: upgrade
+          upgrade: yes
+      - 'JJB-Jenkins-RPC-PR-{type}':
+          type: aio
+          upgrade: no
+
+## Job Definitions
+# This template is for testing PRs against Jenkins-RPC
+# It is triggered by PR and runs an AIO with the proposed version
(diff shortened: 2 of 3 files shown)


git-harry:
@hughsaunders the jobs defined here already exist - how do the two relate? The commit message would be more helpful if it explain the what and the why in more detail.

hughsaunders:
@git-harry the jobs only exist because I manually ran jjb to create them. Any changes made via the jenkins ui will be lost next time jjb is executed (for jobs defined in jjb).

git-harry:
@hughsaunders is the manual step documented anywhere? Is this the standard way of updating JJB jobs? Given that this thing manages its own update job it would be good to have recovery steps somewhere. Maybe the update job should even be in a separate file. My concern is that something gets merged that breaks the job updating and then someone loses a day trying to work out how to fix things.

hughsaunders:
@git-harry scripts/run_jjb.sh (included in this pr) is how I've been manually updating jobs. There is one outstanding problem with this PR, which is that the recheck trigger-phrase isn't working as once in jenkins it has \Q prepended and \E appended. I'm not sure if this a bug with JJB or the github trigger plugin, as JJB seems to write a clean job xml file, but when jenkins updates it, it adds the offending sequences.

sigmavirus24:
So, Jenkins Job Builder is not being used upstream and may fall out of maintenance. Have you considered upgrading to a recentish version of Jenkins and using a Jenkinsfile in RPCO?

odyssey4me:
@sigmavirus24 It appears that JJB is alive and well, according to http://lists.openstack.org/pipermail/openstack-dev/2016-June/097590.html

hughsaunders:
The \Q \E problem has been solved by upgrading the jenkins 2.7.x 🎉 

@sigmavirus24 I was lead to believe that JJB will be maintained even when infra stop using it. I've looked at Jenkinsfile and JJB, and decided that a combination is probably the most powerful. Jenkinsfile is great in combination with the org folder, but it lacks the ability to specify triggers, so periodic jobs can't be created.

I see two possible ways of combining the two options:

- Use JJB to define the structure but use the pipeline DSL for defining the build steps within the JJB jobs.

- Use jenkinsfile & org folder for all the PR jobs and JJB for the periodics. 
  * Pros:
    - Test definitions stored with code
    - Test definitions are branched with code
  * Cons:
    - Job definitions are duplicated in each branch and may need to be updated in multiple places

hughsaunders:
recheck_all

sigmavirus24:
@hughsaunders that sounds like a good way of doing it. I just am very skeptical that an openstack project would continue to be maintained past its usefulness.

hughsaunders:
There is currently an open bug preventing workflow/pipeline jobs from being triggered via the GitHub pull request trigger :(  https://issues.jenkins-ci.org/browse/JENKINS-32556 Pipelines can still be triggered from a PR via an org folder.

hughsaunders:
Note that the above conversation is about JJB, pipelines etc is about the future direction of building jobs in SCM. This discussion doesn't affect this PR which is ready for merge. I will move the discussion to rcbops/u-suk-dev#222.

hughsaunders:
recheck_jjb

sigmavirus24:
This looks good to merge. I'm going to wait for the aio/upgrade builds to fnish though.

sigmavirus24:
Merge-y merge-y make with the yamlification

PR #610, opened 2016-07-06
RPC AIO: Fail Fast

hughsaunders:
This commit adds "set -e" to the aio build script, so that the script
will halt on the first failure that is encountered. Previously the
script continued through failure but stored a binary success/fail value
for each phase. As the script will now halt on failure it is no longer
necessary to store those results. Success lines have been added to the
log to give an indication of the last successfull step when diagnosing
failures.

Connects https://github.com/rcbops/u-suk-dev/issues/223

### DIFF
--- a/scripts/aio_build_script.sh
+++ b/scripts/aio_build_script.sh
@@ -1,4 +1,7 @@
-#!/bin/bash -x
+#!/bin/bash
+set -x # verbose
+set -u # fail on ref before assignment
+set -e # terminate script after any non zero exit
 
 # Useful for getting real time feedback from ansible playbook runs
 export PYTHONUNBUFFERED=1
@@ -10,7 +13,10 @@ log_git_status(){
 }
 
 run_rpc_deploy(){
+  script=${1:-deploy.sh}
+  echo "********************** Run RPC $script ***********************"
   sudo -E\
+    TERM=linux \
     DEPLOY_AIO=yes \
     DEPLOY_HAPROXY=yes \
     DEPLOY_TEMPEST=yes \
@@ -21,25 +27,21 @@ run_rpc_deploy(){
     ANSIBLE_GIT_REPO="https://github.com/hughsaunders/ansible" \
     ADD_NEUTRON_AGENT_CHECKSUM_RULE=yes \
     BOOTSTRAP_OPTS=$BOOTSTRAP_OPTS \
-    scripts/deploy.sh
-}
-
-calc_result(){
-  [ $DEPLOY_RC == 0 -a $TEMPEST_RC == 0 -a $HOLLAND_RC == 0 ]
-  OVERALL_RESULT=$?
-  echo "Deploy Result: $DEPLOY_RC"
-  echo "Tempest Result: $TEMPEST_RC"
-  echo "Holland Result: $HOLLAND_RC"
-  echo "Overall Result: $OVERALL_RESULT"
+    scripts/$script
+  echo "********************** RPC $script Completed Succesfully ***********************"
 }
 
 run_tempest(){
   # jenkins user does not have the necessary permissions to run lxc commands
   # serial needed to ensure all tests
+  echo "********************** Run Tempest ***********************"
   sudo -E lxc-attach -n $(sudo -E lxc-ls |grep utility) -- /bin/bash -c "RUN_TEMPEST_OPTS='--serial' /opt/openstack_tempest_gate.sh ${TEMPEST_TESTS}"
+  echo "********************** Tempest Completed Succesfully ***********************"
 }
 run_holland(){
+  echo "********************** Run Holland  ***********************"
   sudo lxc-attach -n $(sudo lxc-ls |grep galera|head -n1) -- /bin/bash -c "holland bk"
+  echo "********************** Holland Completed Succesfully ***********************"
 }
 
 override_oa(){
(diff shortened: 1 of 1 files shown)


hughsaunders:
recheck

scripts/aio_build_script.sh
```diff
@@ -136,72 +135,37 @@ tempest_tempest_conf_overrides:
...
+      git rebase origin/${ghprbTargetBranch}
     fi
     git submodule sync
     git submodule update --init
     override_oa
     log_git_status
-    echo "********************** Run RPC Deploy/Upgrade Script ***********************"
     if [[ "$UPGRADE_TYPE" == "major" ]]; then
-      sudo -E \
-        TERM=linux \
```
git-harry: This was important for upgrades, not sure why but they always failed without it. This is lost by replacing the command with the function run_rpc_deploy. Has this change been tested?

git-harry:
While I don't see the benefit of running through the tests if the deployment failed. Maybe there is benefit in running all the tests even if one set failed. E.g. still run Holland test if Tempest fail.

git-harry:
@hughsaunders when was that check added? Might be worth adding one for upgrades as well.

scripts/aio_build_script.sh
hughsaunders: Good spot, thanks. Will add an upgrade test.

hughsaunders:
recheck

hughsaunders:
recheck

hughsaunders:
recheck

hughsaunders:
recheck

hughsaunders:
recheck

hughsaunders:
recheck_aio

hughsaunders:
recheck_aio

hughsaunders:
recheck_upgrade

hughsaunders:
recheck_upgrade

scripts/aio_build_script.sh
```diff
@@ -1,4 +1,8 @@
-#!/bin/bash -x
+#!/bin/bash
+TERM=linux \
```
git-harry: Why has this been added?

scripts/aio_build_script.sh
```diff
@@ -172,36 +161,16 @@ if [ "$UPGRADE" == "yes" ] && [ "$OVERALL_RESULT" -eq 0 ];
...
-        BOOTSTRAP_OPTS=$BOOTSTRAP_OPTS \
-        scripts/upgrade.sh
+      run_rpc_deploy upgrade.sh
     else
       run_rpc_deploy
     fi
-    DEPLOY_RC=$?
-
-    echo "********************** Run Tempest ***********************"
     run_tempest
```
git-harry: It might be easier to add another commit at some point define the tests in one place e.g.

```
run_tests(){
  run_tempest
  run_holland
}
```

This would ensure any new tests are called correctly for both greenfields and upgrades.

scripts/aio_build_script.sh
```diff
@@ -140,29 +141,17 @@ tempest_tempest_conf_overrides:
 EOVARS
 
 # Set ubuntu repo to supplied value. Effects Host bootstrap, and container default repo.
-export BOOTSTRAP_OPTS="${BOOTSTRAP_OPTS} bootstrap_host_ubuntu_repo=${UBUNTU_REPO}"
+export BOOTSTRAP_OPTS="${BOOTSTRAP_OPTS:-} bootstrap_host_ubuntu_repo=${UBUNTU_REPO}"
```
git-harry: Why has this been done? I'm not sure how it makes a difference.

hughsaunders: @git-harry with set -u, ref-before-assignment causes the script to exit. If BOOTSTRAP_OPTS is not defined in the environment, then the script would fail. :- sets a default value for the var, effectively defining it at read time. This allows the script to read a variable that may or may not be set, but still fail quickly on any other undefined variables.

scripts/aio_build_script.sh
hughsaunders: Mistake, removed.

scripts/aio_build_script.sh
```diff
@@ -21,25 +27,21 @@ run_rpc_deploy(){
...
-
-calc_result(){
-  [ $DEPLOY_RC == 0 -a $TEMPEST_RC == 0 -a $HOLLAND_RC == 0 ]
-  OVERALL_RESULT=$?
-  echo "Deploy Result: $DEPLOY_RC"
-  echo "Tempest Result: $TEMPEST_RC"
-  echo "Holland Result: $HOLLAND_RC"
-  echo "Overall Result: $OVERALL_RESULT"
+    scripts/$script
+  echo "********************** RPC $script Completed Succesfully ***********************"
```
git-harry: s/Succesfully/Successfully/

scripts/aio_build_script.sh
```diff
@@ -21,25 +27,21 @@ run_rpc_deploy(){
...
 }
 
 run_tempest(){
   # jenkins user does not have the necessary permissions to run lxc commands
   # serial needed to ensure all tests
+  echo "********************** Run Tempest ***********************"
   sudo -E lxc-attach -n $(sudo -E lxc-ls |grep utility) -- /bin/bash -c "RUN_TEMPEST_OPTS='--serial' /opt/openstack_tempest_gate.sh ${TEMPEST_TESTS}"
+  echo "********************** Tempest Completed Succesfully ***********************"
```
git-harry: s/Succesfully/Successfully/

scripts/aio_build_script.sh
```diff
@@ -21,25 +27,21 @@ run_rpc_deploy(){
...
 }
 run_holland(){
+  echo "********************** Run Holland  ***********************"
   sudo lxc-attach -n $(sudo lxc-ls |grep galera|head -n1) -- /bin/bash -c "holland bk"
+  echo "********************** Holland Completed Succesfully ***********************"
```
git-harry: s/Succesfully/Successfully/

git-harry:
👍

evrardjp:
I'm green-buttoning this one.

PR #613, opened 2016-07-14
Fix subshell interaction with -e

hughsaunders:
When -e is set, bash will:
"Exit immediately if a simple command exits with a non-zero status,
unless the subshell is part of the command list immediately following a
`while` or `until` keyword, part of the test in an `if` statement, part
of the right-hand-side of a `&&` or `||` list, or if the command's
return status is being inverted using `!`"

So using -e in a subshell is not reliable especially if the subshell is
part of larger logically combined statement.

This patch removes the reliance on -e in the subshell statements by
replacing `;` with `&&`.

This patch also sets pipefail to catch errors that may have been masked
by pipes.

Related: rcbops/u-suk-dev#223

### DIFF
--- a/scripts/aio_build_script.sh
+++ b/scripts/aio_build_script.sh
@@ -2,6 +2,7 @@
 set -x # verbose
 set -u # fail on ref before assignment
 set -e # terminate script after any non zero exit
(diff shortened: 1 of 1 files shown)


sigmavirus24:
👍

PR #616, opened 2016-07-15
Keeupundefinedparamters

hughsaunders:
Jenkins change SECURITY-170 introduces a change that prevents jobs from
receving undefined parameters into their environment.

The current RPC job set relies on parameters from the github pull
request trigger, some of which are not defined as job parameters. This
patch sets jenkins back to the old behaviour, which gives us time to
change the jobs inline with the new standard.

### DIFF
--- a/dockerfiles/jenkins/Dockerfile
+++ b/dockerfiles/jenkins/Dockerfile
@@ -35,4 +35,4 @@ ENV JENKINS_HOME /opt/jenkins
 VOLUME /opt/jenkins
 
 # Fire up jenkins
-CMD java -jar /opt/jenkins/jenkins.war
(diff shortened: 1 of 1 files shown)


hughsaunders:
recheck_all

hughsaunders:
see also: https://jenkins.io/blog/2016/05/11/security-update/

evrardjp:
Could you add this future change (and the revert to new standard "keepundefinedparameters false") to the gating epic?

hughsaunders:
Something else to note... The docker image that the jenkins master is currently running from already has this change. This PR is just to ensure the repo matches the running code. 

@evrardjp I added a reference to this in rcbops/u-suk-dev#222

hughsaunders:
recheck_aio

hughsaunders:
previous fail was timeout trying to ssh to horizon container

evrardjp:
👍

PR #617, opened 2016-07-19
Fix the horizon extensions override

evrardjp:
The horizon extension override in the user_variables doesn't
override the subfolder. Due to the way the py_pkg lookup works,
all the variables that need to be overriden have to be specified
in the user_variables file.


hughsaunders:
👍

PR #621, opened 2016-07-25
Add multinode cleanup script.

hughsaunders:
This script is currently part of the unversioned Jenkins config for the
job http://jenkins.propter.net/job/RPC-Multinode-Cleanup/configure.

As well as importing the script into this repo, I have also added a
section for cleaning up networks. This is because heat sometimes fails
to delete networks, so they have been accumulating and we recently hit
the test account network quota.

Connects rcbops/u-suk-dev#244

### DIFF
--- a/scripts/heat_cleanup.sh
+++ b/scripts/heat_cleanup.sh
@@ -36,3 +36,46 @@ while read stack_ref; do
                 done
     fi
 done <stack_refs
+
+# Delete all networks that are not associated with an active stack.
+
+# I failed to get rackspace-novaclient to interface with rackspace networks.
+# Shade won't either, so I resorted to using pyrax. However pyrax doesn't
+# support orchestation, so I'm using shade for that.
+python <<EOP
+import os
+import os_client_config
+import pyrax
+import shade
+
+# Note that clouds.yaml and mattt_jenkins_dfw.pyrax contain the same creds
+# but in different formats, and must be updated in unison.
+
+# configure shade with creds from clouds.yaml
+cloud_config = os_client_config.OpenStackConfig(
+  config_files=['/opt/jenkins/creds/clouds.yaml']
+).get_one_cloud('raxus', region_name='DFW')
+osc = shade.OpenStackCloud(cloud_config=cloud_config)
+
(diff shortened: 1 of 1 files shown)


hughsaunders:
ignore the CI, it won't test this script.

hughsaunders:
recheck_aio

scripts/heat_cleanup.sh
```diff
@@ -0,0 +1,76 @@
...
+# I failed to get rackspace-novaclient to interface with rackspace networks.
+# Shade won't either, so I resorted to using pyrax. However pyrax doesn't
+# support orchestation, so I'm using shade for that.
+python <<EOP
+import os
+import os_client_config
+import pyrax
+import shade
+
+# configure shade with creds from clouds.yaml
```
git-harry: Is the fact that clouds.yaml and mattt_jenkins_dfw.pyrax contain the same thing documented anywhere? Maybe clouds.yaml could have a similar name to make it clearer. Not a big deal but this is probably going to cause a breakage down the line when one of those files is updated but not the other.

scripts/heat_cleanup.sh
```diff
@@ -0,0 +1,76 @@
...
+while read stack_ref; do
+    if ! grep -q $stack_ref stacks_to_keep; then
+        cinder --os-volume-api-version 1 list 2>/dev/null\
+            |awk '/'$stack_ref'/{print $2}' \
+                |while read vol_id; do
+                    cinder --os-volume-api-version 1 delete $vol_id 2>/dev/null
+                done
+    fi
+done <stack_refs
+
```
git-harry: similar to the other sections it probably would be nice to have had a comment here along the lines of:

```
# Delete all networks that are associated with stacks that should be removed
```

git-harry:
👍 couple of minor points. It also would have been nice if the commit to move the code to the repo was separate from the one to change it.

scripts/heat_cleanup.sh
hughsaunders: This is on L67, but will add it above the entire section as well.

git-harry:
👍

PR #624, opened 2016-07-28
Fix the horizon extensions override (redux)

evrardjp:
We forgot to add the git package name, so it's considered as
another egg for the py_pkg lookup.

This should fix it.

Connects rcbops/u-suk-dev#259


hughsaunders:
👍

cloader89:
LGTM. 👍

PR #627, opened 2016-08-11
Increase interval between ssh connections attempts

hughsaunders:
This patch overrides the wait for ssh delay time. This will give
containers longer to start before failing on ssh.

Connects rcbops/u-suk-dev#279


hughsaunders:
This may not be the solution as some of the builds are getting timed out by jenkins after 20 minutes. 

Current settings are:
```
Delay: 5
Retries: 3
wait_for timeout: 300 (Default)
```

Maximum possible amount of time = 300*3 + 5*2 = 910 seconds (~15.2 mins) So how does 20 minutes elapse without console update?

hughsaunders:
Just checked, of the container ssh related failures in the last 30 days, 15 were build timeouts, 20 were too many-retries. There haven't been any build timeouts it the last couple of days, this may be related to the switch to JJB defined jobs, which included an inactivity timeout increase from 20 mins to 30 mins. 

So increasing the retries may still be a good strategy for reducing the failures that were due to running out of retries.

git-harry:
What are your thoughts in putting this in RPCO instead?

hughsaunders:
@git-harry fine by me. Probably save some hassle in non-aio situations.

hughsaunders:
WIP While I figure out where this needs to go in RPC.

hughsaunders:
Closing - wip comment was meant for waffle issue, not the PR.

PR #629, opened 2016-08-15
Write to gating vars file as root

hughsaunders:
Resolves Permission Denied error involving user_zzz_gating_variables.yml.

Connects rcbops/u-suk-dev#284


git-harry:
👎 as discussed, this should not be required. The permissions failure happens because the file exists when it shouldn't.

PR #632, opened 2016-08-15
Remove unnecessary BOOTSTRAP_OPTS pass throughs

hughsaunders:
These lines are not necessary as sudo -E will preserve the envionment.
The BOOTSTRAP_OPTS line was preventing the BOOTSTRAP_OPTS var from being
passed from jenkins-rpc/aio_build_script to
rpc-openstack/scripts/deploy.sh because it had a leading space but
wasn't quoted.

Connects rcbops/u-suk-dev#263


git-harry:
I think it would be better not to use 'sudo -E', while in some respects it may make managing new variables easier, explicitly defining the variables that are used makes it easier to see what the script is doing. That said, given this method is already being used bringing all the variables into line does not seem unreasonable.

PR #634, opened 2016-08-16
Increase ping thread timeout

hughsaunders:
This may solve the problem with instances being killed during upgrade
jobs before the job is complete.

Also adds libssl-dev to the jenkins master docker image as it is needed
for building the cryptography package.

Connects rcbops/u-suk-dev#302

### DIFF
--- a/dockerfiles/jenkins/Dockerfile
+++ b/dockerfiles/jenkins/Dockerfile
@@ -21,7 +21,7 @@ RUN apt-get update
 
 # Upgrade and install things
 RUN apt-get upgrade -y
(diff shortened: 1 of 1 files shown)


git-harry:
@hughsaunders what's the cryptography thing?

hughsaunders:
@git-harry One of the dockerfile steps is to install the following python packages: "requests pyyaml virtualenv "ansible<2.0" jinja2 python-heatclient" one of which depends on cryptography. Cryptography requires openssl c libraries, so the libssl-dev package has been added. 

The python packages are not versioned, so the cryptography dep could have been bought in via a new version of one of the packages in the chain.

git-harry:
@hughsaunders okay makes sense, thanks.

PR #637, opened 2016-08-17
[WIP] Catch exceptions when reading jobs

hughsaunders:
Jobs may have incomplete logs/configs if they are aborted at a bad time.
Build summary should not fail if one log is unreadable, it should skip
and continue.

This patch adds primative exception handling and logging to show which
jobs have files that cannot be read. Note that logging is to stderr
because the html is written to stdout.

Connects rcbops/u-suk-dev#311


hughsaunders:
Marking as WIP and will re do with Harry's suggestions:
> git-harry:
 hughsaunders: I just thought it might be easier to keep them all in one place, move the error   handling closer to where it happens and add the failures to the build summary page. That way all   the information is in one place and if there is a trend we are more likely to spot it.

hughsaunders:
Having looked into this, I've changed my mind. Accepting any build XML file with any missing element breaks lots of assumptions that the build summary scripts make. For example, if the timestamp element is missing, it will break the UI when that value is piped through the date formatting filter. It would be messy to put conditionals in for every element access.

I think skipping and logging invalid build files, is cleaner. Build summary isn't a monitoring system - it's just an overview.

hughsaunders:
This PR can't be reopend 😠

PR #638, opened 2016-08-17
Catch exceptions when reading jobs

hughsaunders:
Jobs may have incomplete logs/configs if they are aborted at a bad time.
Build summary should not fail if one log is unreadable, it should skip
and continue.

This patch adds primative exception handling and logging to show which
jobs have files that cannot be read. Note that logging is to stderr
because the html is written to stdout.

Connects rcbops/u-suk-dev#311
Continuation of #637

### DIFF
--- a/scripts/build-summary/build-summary-gh.py
+++ b/scripts/build-summary/build-summary-gh.py
@@ -7,6 +7,8 @@ import datetime
 import os
 import pickle
 import re
+import sys
+import traceback
 
 # 3rd Party imports
 import click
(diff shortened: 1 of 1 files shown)


hughsaunders:
Relevant comments from un-reponable pr #637

1) 
> Marking as WIP and will re do with Harry's suggestions:
>>git-harry:
hughsaunders: I just thought it might be easier to keep them all in one place, move the error handling closer to where it happens and add the failures to the build summary page. That way all the information is in one place and if there is a trend we are more likely to spot it.

2)
>Having looked into this, I've changed my mind. Accepting any build XML file with any missing element breaks lots of assumptions that the build summary scripts make. For example, if the timestamp element is missing, it will break the UI when that value is piped through the date formatting filter. It would be messy to put conditionals in for every element access.

>I think skipping and logging invalid build files, is cleaner. Build summary isn't a monitoring system - it's just an overview.

hughsaunders:
Note that I have left traceback prints in the debug output, so if some failures would be better handled internally the console log should give enough information to do so.

hughsaunders:
Example of incomplete xml file: https://gist.github.com/hughsaunders/4648fdd26436da2ab53064d5d5694a10

mancdaz:
👍  after discussion with @hughsaunders I think this makes sense.

PR #639, opened 2016-08-19
Define RPC-AIO PR & Periodic jobs in JJB

hughsaunders:
This PR adds JJB Job definitions for RPC-AIO jobs including:
  * 8 RPC-AIO jobs (kilo,liberty,mitaka,master)x(swift,ceph). These jobs are triggered by PRs and a schedule. So they replace the existing -PR jobs, and the Periodic-Matrix. 
  * A Major upgrade job. This is also dual triggered by PR to liberty and schedule. This does not replace the upgrade matrix, that will be addressed in a subsequent PR. 
 * A job to test JJB job definitions in pull requests. 

These jobs have been running as periodics for a few days with good results, once this PR is merged and JJB is rerun, these jobs will also run against PRs. 

Connects rcbops/u-suk-dev#222

### DIFF
--- a/rpc-jobs/jobs.yaml
+++ b/rpc-jobs/jobs.yaml
@@ -1,4 +1,4 @@
-## Macros for repeated blocks
+# Macros for repeated blocks
 
 # Define the github project associated with jenkins rpc
 - property:
@@ -6,6 +6,11 @@
     properties:
       - github:
           url: https://github.com/rcbops/jenkins-rpc
+- property:
+    name: rpc-openstack-github
+    properties:
+      - github:
+          url: https://github.com/rcbops/rpc-openstack
 
 # Define the scm/git repo associated with jenkins-rpc
 - scm:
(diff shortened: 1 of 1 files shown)


rpc-jobs/jobs.yaml
```diff
@@ -18,22 +23,311 @@
...
+            popd
+            ./buildscript_repo/$BUILD_SCRIPT_PATH
+          }}
+
+          prep_artefacts(){{
+            echo "****** Copy Logs To Jenkins Workspace For Archival *******"
+
+            #copy logs into jenkins workspace so they can be archived with the results
+            mkdir -p archive
+            cp -rL /openstack/log archive/openstack
```
odyssey4me: Do you need to be concerned with whether the log files (or any other gathered artefacts) are not readable? If jenkins runs as non-root this may be an issue.

rpc-jobs/jobs.yaml
```diff
@@ -18,22 +23,311 @@
...
+            cp -r /var/lib/lxc/*galera*/rootfs/etc/mysql archive/etc/
+            cp -r /var/lib/lxc/*memcached*/rootfs/etc/memcached.conf archive/etc/
+
+            # Collect MAAS agent config
+            cp -r /etc/rackspace-monitoring-agent.conf.d archive/etc/
+            cp /etc/rackspace-monitoring-agent.cfg archive/etc/
+
+            # Jenkins user must be able to read all files to archive and transfer to
+            # the master. Unreadable files will cause the whole archive operation to fail
+            chown -R jenkins archive
```
odyssey4me: Ah, never mind - I see you have taken care of this.

odyssey4me:
I don't know enough about how Jenkins is setup, but I don't see anything horrible or scary here. This looks good to me.

rpc-jobs/jobs.yaml
```diff
@@ -18,22 +23,311 @@
...
+          status-context: '{context}'
+    scm:
+      - git:
+          url: "{RPC_REPO}"
+          branches:
+            - "${{sha1}}"
+          refspec: "+refs/pull/*:refs/remotes/origin/pr/* +refs/heads/*:refs/remotes/origin/*"
+          name: origin
+    builders:
+      - shell: |
```
git-harry: Is it possible to reference to have this as a separate script and reference it by name? It allows for syntax highlighting which makes it easier to read.

hughsaunders: This script will be integrated with aio_build_script, and probably a separate file for a functions lib, but I would like to put that refactor in its own change.

hughsaunders:
recheck_all - the aio jobs don't test this patch, but failed builds put people off merging.

hughsaunders:
recheck_upgrade

andymcc:
I agree with Jesse, I don't see anything breaking or terrible, I imagine its quite hard to get solid reviews when you are the main/only expert though. Happy for this to merge.

PR #641, opened 2016-08-22
Match Tracebacks

hughsaunders:
Python tracebacks are indicative of a problem, they should be logged in
build summary.

This is a generalisation of a previous filter which only matched tempest
exceptions.

Connects rcbops/u-suk-dev#342


hughsaunders:
closing to prevent excessive PR testing will make this more efficient

PR #642, opened 2016-08-22
Generalise APT Failures

hughsaunders:
Catch apt failures other than hash sum mismatch. This will help track
404s and other mirror related failures.

Connects rcbops/u-suk-dev#346

### DIFF
--- a/scripts/build-summary/build.py
+++ b/scripts/build-summary/build.py
@@ -205,11 +205,11 @@ class Build(object):
                                      fail=match.group(1)))
 
     def apt_fail(self, lines):
-        match_re = re.compile("W: Failed to fetch (.*) *Hash Sum mismatch")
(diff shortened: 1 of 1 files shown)


odyssey4me:
@hughsaunders You may wish to split the failures into two sets:

1. apt-get install / apt-get update failures
2. apt key-related failures

hughsaunders:
@odyssey4me thanks for looking. This generalised pattern doesn't catch key failures. They are currently caught by a different pattern that catches "too many retries" on an ansible task. It may be useful to add one that tracks keys explicitly, however they are trackable currently:
screen shot 2016-08-22 at 11 37 03

odyssey4me:
Ah, thanks @hughsaunders - I just wanted to make sure that both options were captured separately. Good show - LGTM! 👍

PR #644, opened 2016-08-24
Match Tracebacks

hughsaunders:
Python tracebacks are indicative of a problem, they should be logged in
build summary.

This is a generalisation of a previous filter which only matched tempest
exceptions.

Also:
  * Refactors the way filters are called so that slow filters can be logged
  * Refactors the failure normalisation function
  * Returns N/A for previous task when the failure occurs outside
    after a successfull ansible run.

Connects rcbops/u-suk-dev#342

### DIFF
--- a/scripts/build-summary/build.py
+++ b/scripts/build-summary/build.py
@@ -1,11 +1,16 @@
 # Stdlib import
 import datetime
 import re
+import sys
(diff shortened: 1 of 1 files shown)


andymcc:
👍

PR #645, opened 2016-08-24
Track dhcp and sshd cirros startup failures

hughsaunders:
Tracking these issues will help us to work out why ssh connections to
instances within tempest tests fail.

Connects rcbops/u-suk-dev#355

### DIFF
--- a/scripts/build-summary/build.py
+++ b/scripts/build-summary/build.py
@@ -188,6 +188,8 @@ class Build(object):
             self.apt_fail,
             self.holland_fail,
             self.slave_died,
+            self.cirros_dhcp,
+            self.cirros_sshd,
 
             # Heat related failures
             self.create_fail,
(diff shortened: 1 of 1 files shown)


git-harry:
Do you have an example log(s)?

hughsaunders:
@git-harry  http://jenkins.propter.net/job/JJB-RPC-AIO_liberty-upgrade/35/consoleFull and http://jenkins.propter.net/job/JJB-RPC-AIO_master-ceph/40/consoleFull Those are from running build summary with this patch applied.

git-harry:
Both logs contain both patterns, do they always go together? Are you attempting to determine the overlap between the two?

hughsaunders:
@git-harry I want to know if they relate to tempest failures such as reboot_server_hard and basic_ops. 

They may always happen together, I'm not sure yet. There definitely could be sshd start failures that aren't related to dhcp.

PR #646, opened 2016-08-25
Fix cuase hierachy for single level jobs

hughsaunders:
Previously build summary assumed that AIO Jobs have an upstream cause,
and the cause hierachy was constructed by iterating over all child
upstreamcauses elements.

This fails to construct an accurate hierachy on JJB jobs as they are
triggered directly rather than by an upstream job.

Connects rcbops/u-suk-dev#365


hughsaunders:
PR / Commit connected to the wrong issue, closing.

PR #647, opened 2016-08-25
Fix cause hierachy for single level jobs

hughsaunders:
Fix cause hierachy for single level jobs
Previously build summary assumed that AIO Jobs have an upstream cause,
and the cause hierachy was constructed by iterating over all child
upstreamcauses elements.

This fails to construct an accurate hierachy on JJB jobs as they are
triggered directly rather than by an upstream job.

This commit also adds recognition for timer triggers and manually
started jobs.

Connects rcbops/u-suk-dev#364

### DIFF
--- a/scripts/build-summary/build.py
+++ b/scripts/build-summary/build.py
@@ -77,35 +77,72 @@ class Build(object):
         jenkins_base = "http://jenkins.propter.net/"
         self.trigger = "periodic"
         self.build_hierachy = []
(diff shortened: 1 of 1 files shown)


git-harry:
@hughsaunders do you have a test page showing what this does?

hughsaunders:
@git-harry as discussed: http://jenkins.propter.net:443/index_test.html

PR #650, opened 2016-09-19
Move artefact archival into post build script

hughsaunders:
Move artefact prep into a post build action. This means that artefact
prep will still be executed if the job is halted by a timeout.

Connects rcbops/u-suk-dev#416


hughsaunders:
recheck_all

PR #656, opened 2016-09-27
Ensure proper variable values for NEW DEFAULTS.

alextricity25:
This is because of the way migrate-yaml.py behaves with the
'--for-testing-take-new-vars-only'; meaning that the new
default variables will be set in the user_*_variables_overrides.yml
file. Since raw_multi_journal is set to False as part of the deploy.sh
process, but is set to True in Mitaka's
user_rpco_user_variables_defaults.yml file, this will result in
migrate-yaml.py adding 'raw_multi_journal: True' in the overrides.
To avoid this behavior in gate, it is overridden here.
The same is true for journal_size, and maas_notification_plan.
They need to be set to their AIO values, and not take
on the new values migrate-yaml.py has given them.

Connects https://github.com/rcbops/u-suk-dev/issues/383

### DIFF
--- a/scripts/aio_build_script.sh
+++ b/scripts/aio_build_script.sh
@@ -189,6 +189,23 @@ tempest_tempest_conf_overrides:
     snapshot: True
 EOVARS
 
+# Ensure raw_multi_journal is False for upgrades.
+# This is because of the way migrate-yaml.py behaves with the
+# '--for-testing-take-new-vars-only'; meaning that the new
+# default variables will be set in the user_*_variables_overrides.yml
+# file. Since raw_multi_journal is set to False as part of the deploy.sh
+# process, but is set to True in Mitaka's
+# user_rpco_user_variables_defaults.yml file, this will result in
+# migrate-yaml.py adding 'raw_multi_journal: True' in the overrides.
+# To avoid this behavior in gate, it is overridden here.
+# The same is true for journal_size, and maas_notification_plan.
+if [ "$UPGRADE" == "yes" ]; then
+    echo "raw_multi_journal: false" | tee -a $uev
+    echo "journal_size: 1024" | tee -a $uev
+    echo "maas_notification_plan: npTechnicalContactsEmail" | tee -a $uev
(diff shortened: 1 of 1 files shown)


mattt416:
👍 pending gate pass

evrardjp:
Should we force osd_directory here too, just in case?

alextricity25:
``osd_directory`` is set by ``deploy.sh`` and it isn't defined as a default in the default files. Thus, ``migrate-yaml.py`` won't see it as a "NEW_DEFAULT" and it's value will remain to be ``True``.

scripts/aio_build_script.sh
```diff
@@ -189,6 +189,20 @@ tempest_tempest_conf_overrides:
...
+# This is because of the way migrate-yaml.py behaves with the
+# '--for-testing-take-new-vars-only'; meaning that the new
+# default variables will be set in the user_*_variables_overrides.yml
+# file. Since raw_multi_journal is set to False as part of the deploy.sh
+# process, but is set to True in Mitaka's
+# user_rpco_user_variables_defaults.yml file, this will result in
+# migrate-yaml.py adding 'raw_multi_journal: True' in the overrides.
+# To avoid this behavior in gate, it is overridden here.
+# The same is true for journal_size, and maas_notification_plan.
+echo "raw_multi_journal: false" | tee -a $uev
```
sigmavirus24: If these are only for upgrades, should these be in a conditional that checks if we're running an upgrade first?

alextricity25: Good point.
These variables are already set to these values on greenfield AIOs. They are set in the deploy.sh script(i.e [maas_notification_plan](https://github.com/rcbops/rpc-openstack/blob/master/scripts/deploy.sh#L75)). Having them defined a second time int he ``user_zzz``` file won't change the way greenfields behave.
I do agree that having the conditional would clear things up, and could possibly avoid any nonresistances if the values in ``deploy.sh`` change.

alextricity25:
I've added ``osd_directory: true`` and a conditional that checks if the build is an upgrade to the latest commit.

Issue #657, opened 2016-10-03
verify-maas.yml needs more time after an upgrade


alextricity25:
``verify-maas.yml`` needs more time for checks and alarms to come up before it stops trying.
I propose we increase the number of ``retries`` and ``delay`` for the tasks in ``verify-maas.yml``

git-harry:
@alextricity25 this work is not being properly tracked, we currently use swimlanes to categories cards so that the appropriate team can focus on them.

Regarding the issue itself, I'm not convinced that this is solving anything. It looks like we're decreased timeouts needlessly and now we're increasing them without getting to the root cause of the issue.

/cc @b3rnard0

alextricity25:
@git-harry:  The root of the issue is that the MaaS system isn't as responsive as we would like it to be. I could be wrong, but I don't think that's something we can control within our scope of work.
Where would be a good place for this issue so we can properly track it? ``u-suk-dev``?

alextricity25:
The *best* thing to do would be to get https://github.com/rcbops/rpc-openstack/pull/1445 merged.

alextricity25:
Closing in favor of rcbops/rpc-openstack#1445

PR #658, opened 2016-10-03
Increase retries and delay for verify-maas tasks

alextricity25:
This PR overrides the ``reties`` and ``delay`` directive
for the ``verify-maas.yml`` tasks because checks and
alarms take a little longer to come up after an upgrade.

Connects #657


alextricity25:
This shouldn't be needed after: https://github.com/rcbops/u-suk-dev/issues/481

PR #660, opened 2016-10-05
Back to rackspace mirror

hughsaunders:
Further work is required before the infra mirrors can be used because
they don't carry all the sections of the ubuntu repo.

Connects rcbops/u-suk-dev#482

### DIFF
--- a/rpc-jobs/jobs.yaml
+++ b/rpc-jobs/jobs.yaml
@@ -113,7 +113,7 @@
     UPGRADE_FROM_REF: "origin/kilo"
     UPGRADE_FROM_NEAREST_TAG: "yes"
     UPGRADE_TYPE: "major"
-    UBUNTU_REPO: "auto"
+    UBUNTU_REPO: "http://mirror.rackspace.com/ubuntu"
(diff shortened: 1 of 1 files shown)


rpc-jobs/jobs.yaml
```diff
@@ -113,7 +113,7 @@
     UPGRADE_FROM_REF: "origin/kilo"
     UPGRADE_FROM_NEAREST_TAG: "yes"
     UPGRADE_TYPE: "major"
-    UBUNTU_REPO: "auto"
+    UBUNTU_REPO: "https://mirror.rackspace.com/ubuntu"
```
odyssey4me: Are you sure you want HTTPS? That may break some things. The value ``http://mirror.rackspace.com/ubuntu/`` will work too.

odyssey4me:
👍 LGTM

rpc-jobs/jobs.yaml
hughsaunders: @odyssey4me I used https as that was the value that was the previous value, have changed it to http for now.

PR #663, opened 2016-10-07
Add set-home option to sudo

hughsaunders:
Our playbooks assume $HOME is set to /root, which may not be the case
with normal sudo. Adding -H sets $HOME to that of the target user.

Connects rcbops/u-suk-dev#488


alextricity25:
:+1: pending checks

PR #664, opened 2016-10-07
Add graphs for upgrade and mitaka builds

hughsaunders:
Currently these are lumped in with build types, but they should have
their own trends for reliability monitoring.

Connects rcbops/u-suk-dev#466


hughsaunders:
Test page generated with code in this PR: http://jenkins.propter.net:443/index_test.html

PR #665, opened 2016-10-10
Add series parameter to build object

hughsaunders:
This will allow us to track builds by series rather than by branch.
This allows us to track all of mitaka rather than only mitaka-13.1 for
example.

Connects rcbops/u-suk-dev#495

### DIFF
--- a/scripts/build-summary/cachequery.py
+++ b/scripts/build-summary/cachequery.py
@@ -1,6 +1,8 @@
 import click
 import pickle
 
+from build import Build
+
 
 @click.group()
 def cli():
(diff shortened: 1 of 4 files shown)


hughsaunders:
Screen shot of the [test page](http://jenkins.propter.net:443/index_test.html) incase the test page is repurposed.
screen shot 2016-10-10 at 13 49 44

evrardjp:
Else the rest seems good and makes sense.

scripts/build-summary/cachequery.py
```diff
@@ -14,10 +16,10 @@ def query(cache_file, query):
     with open(cache_file, 'rb') as f:
         key, criteria = query.split('=')
         buildobjs = pickle.load(f)
-        for build in buildobjs:
-            item = build.getattr(key)
+        for name, build in buildobjs.items():
```
hughsaunders: Correct no, it isn't. I was using it in the final print, but then fixed the str() function on build objects, so didn't need the name anymore. Not worth revising just to change name to _ though. Especially as it's in a script that I only use for debugging.

PR #667, opened 2016-10-10
Enable colours in logs

hughsaunders:
Ansible is outputing colour control codes, but jenkins isn't
interpreting them, so the logs are littered with non printing characters
which reduces readability.

This patch enables the jenkins plugin for interpeting colour control codes.

Connects rcbops/u-suk-dev#361


d34dh0r53:
:+1:

PR #668, opened 2016-10-17
Store ceph.conf as artifact

alextricity25:
This commit follows the same pattern for openstack configurations
to store ceph.conf configurations. It uses the find command to search
for ceph configurations inside container filesystems, and as well as
the host filesystem. It then copies said configurations into the
the appropriate directory.

It also makes some small formatting changes to code comments.

Connects https://github.com/rcbops/u-suk-dev/issues/373


alextricity25:
recheck_upgrade

mattt416:
recheck_upgrade

mattt416:
This LGTM, but need to see upgrade test passing.

PR #671, opened 2016-11-04
Move misc jobs to JJB

mattt416:
This commit moves the following jobs defined manually in Jenkins to JJB:

- RPC-AIO Build Summary
- RPC-AIO Instance Cleanup
- RPC-Multinode-Cleanup

We also create a new file called scripts/aio_cleanup.py which is called
by one of the JJB jobs to avoid embedding this rather lenghty code into
JJB itself.

Connects https://github.com/rcbops/u-suk-dev/issues/352

### DIFF
--- a/rpc-jobs/jobs.yaml
+++ b/rpc-jobs/jobs.yaml
@@ -113,6 +113,37 @@
           context: cidev
           CRON: ""
 
+- project:
+    name: 'JJB-Misc-Jobs'
+    jobs:
+      - 'JJB-Misc-{name}':
+          name: Build-Summary
+          shell: |
+            #!/bin/bash -x
+            . /opt/jenkins/venvs/buildsummary/bin/activate
+            cd scripts/build-summary
+            python build-summary-gh.py /opt/jenkins_builds/jobs/{{RPC-AIO,RPC-Training-Multinode,JJB-RPC-AIO*,JJB-Jenkins-RPC-PR*}}/builds/*/build.xml \
+                > /opt/jenkins/www/index_tmp.html \
+                && cp /opt/jenkins/www/index_tmp.html /opt/jenkins/www/index.html
+      - 'JJB-Misc-{name}':
+          name: AIO-Cleanup
+          shell: |
+            #!/bin/bash
+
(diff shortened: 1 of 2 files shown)


rpc-jobs/jobs.yaml
```diff
@@ -113,6 +113,37 @@
...
+- project:
+    name: 'JJB-Misc-Jobs'
+    jobs:
+      - 'JJB-Misc-{name}':
+          name: Build-Summary
+          shell: |
+            #!/bin/bash -x
+            . /opt/jenkins/venvs/buildsummary/bin/activate
+            cd scripts/build-summary
+            python build-summary-gh.py /opt/jenkins_builds/jobs/{RPC-AIO,RPC-Training-Multinode,JJB-RPC-AIO*,JJB-Jenkins-RPC-PR*}/builds/*/build.xml \
```
hughsaunders: {foo} is used to expand jjb variables, so bash braces need to be doubled. 
```python build-summary-gh.py /opt/jenkins_builds/jobs/{{RPC-AIO,RPC-Training-Multinode,JJB-RPC-AIO*,JJB-Jenkins-RPC-PR*}}/builds/*/build.xml```

**Comment by mattt416** [comment]
To anyone reviewing -- we will need to merge this in a controlled environment as we don't want potentially broken JJB jobs to be rolled out without anyone paying much attention.

File: rpc-jobs/jobs.yaml [hunk 2]
```diff
@@ -498,6 +529,28 @@
...
+    description: 'Managed by JJB: Misc {name}'
+    defaults: global
+    disabled: false
+    concurrent: false
+    node: master
+    logrotate:
+      numToKeep: 50
+    # NOTE(mattt): Is this property actually required here?
+    properties:
+      - jenkins-rpc-github
```
hughsaunders: I don't think this is needed, as none of the jobs in this project are triggered by PR.

hughsaunders:
I've disabled the jobs that these jobs replace and will merge this shortly, I will then monitor the initial runs of the new jobs.

hughsaunders:
I have checked the output of each of the generated jobs and they all look good, thanks @mattt416

PR #672, opened 2016-11-07
Move logstash plugin to use post-build component

alextricity25:
This commit removes the logstash buildwrapper and adds the
logstash post-build publisher so logs are processed after
a build completes, and not on a line-by-line basis.

Connects https://github.com/rcbops/u-suk-dev/issues/540

### DIFF
--- a/rpc-jobs/jobs.yaml
+++ b/rpc-jobs/jobs.yaml
@@ -156,8 +156,6 @@
     logrotate:
       daysToKeep: 30
     wrappers:
-      - logstash:
-          use-redis: False
       - ansicolor
       - timestamps
       - credentials-binding:
@@ -459,6 +457,10 @@
       - junit:
           results: "archive/openstack/*_utility_*/*.xml"
           allow-empty-results: true
+      # Push logs over to logstash
+      - logstash:
+          max-lines: 70000
+          fail-build: false
 
 # This template has two variables - type and upgrade.
 - job-template:


rpc-jobs/jobs.yaml
```diff
@@ -459,6 +457,10 @@
       - junit:
           results: "archive/openstack/*_utility_*/*.xml"
           allow-empty-results: true
+      # Push logs over to logstash
+      - logstash:
+          max-lines: 2000
```
mattt416: Why 2000?

alextricity25: That number isn't based on anything. I just figured 2000 lines of the tailed log would be enough for us to capture any errors. Do you think it's enough?

mattt416: Ah, I see.  I think it'd be nice if we just sent everything across tbh.

alextricity25: I personally don't see the need for all of the logs to be sent over as one record. So long as we have enough lines to capture the output, we should be okay. Also, keep in mind that the Jenkins Logstash plugin will also construct fields based on job meta-data. Looking at the example payload [here](https://wiki.jenkins-ci.org/display/JENKINS/Logstash+Plugin#LogstashPlugin-JSONPayloadFormat), we should have everything we need even though we are just tailing 2000 lines of the build's log.

hughsaunders: I'd prefer to have the entire console, its often useful to know what happened previously when debugging an error, eg "was the task thats supposed to install package x skipped?"

I looked at a selection of AIO console logs and they were all <30k lines, so an upgrade should be <60k, add 10k for margin and end up with a limit of 70k? 

I'm not sure if that will cause any of the components to explode, but worth a try 😎  also probably more efficient than 70k one line records.

mattt416:
👍 Hold onto your butts!

PR #673, opened 2016-11-17
Add JJB definition for heat multinode job

hughsaunders:
Replaces the UI defined RPC-Training-Multinode job with
JJB-RPC-Training-Multinode.

Connects rcbops/u-suk-dev#351


hughsaunders:
Job is already running: http://jenkins.propter.net/job/JJB-RPC-Training-Multinode/
There are some failures, but they are tempest failures, not JJB related.

PR #674, opened 2016-11-17
Add cirros_tgz_url variable

alextricity25:
The cirros_tgz_url variable has been dropped in OSA newton, but is still
used by earlier releases that we test. This commit adds the variable in the
variable overrides file so the newton tempest role can complete succesfully.

Related https://github.com/rcbops/rpc-openstack/pull/1577
Connects https://github.com/rcbops/u-suk-dev/issues/618

### DIFF
--- a/scripts/aio_build_script.sh
+++ b/scripts/aio_build_script.sh
@@ -178,6 +178,9 @@ fi
 
 #Supply fixed cirros image while empty key bug is not fixed upstream.
 tee -a $uev >/dev/null <<EOVARS
+# (alextricity25) Adding the cirros_*_url vars. This var has been dropped from OSA in
+# newton so we are carrying it here so we have support for both newton and earlier releases.
+cirros_tgz_url: "http://download.cirros-cloud.net/{{ cirros_version }}/cirros-{{ cirros_version }}-x86_64-uec.tar.gz"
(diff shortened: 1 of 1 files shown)


scripts/aio_build_script.sh
```diff
@@ -178,6 +178,9 @@ fi
 
 #Supply fixed cirros image while empty key bug is not fixed upstream.
 tee -a $uev >/dev/null <<EOVARS
+# (alextricity25) Adding the cirros_*_url vars. This var has been dropped from OSA in
+# newton so we are carring it here so we have support for both newton and earlier releases.
+cirros_tgz_url: "http://download.cirros-cloud.net/{{ cirros_version }}/cirros-{{ cirros_version }}-x86_64-uec.tar.gz"
```
odyssey4me: The reason this was removed for newton is because tempest deprecated it, so I think something else is going to have to be done here. Simply adding this doesn't seem like it'll achieve the right results.

scripts/aio_build_script.sh
```diff
@@ -178,6 +178,9 @@ fi
 
 #Supply fixed cirros image while empty key bug is not fixed upstream.
 tee -a $uev >/dev/null <<EOVARS
+# (alextricity25) Adding the cirros_*_url vars. This var has been dropped from OSA in
+# newton so we are carring it here so we have support for both newton and earlier releases.
```
nrb: carrying*

d34dh0r53:
👍

PR #675, opened 2016-11-22
Copy horizon error.log files into jenkins archive

mattt416:
This commit copies the horizon container's error.log file into the
jenkins archive.  This will hopefully help us identify the cause of the
apache2 restart failures that we're seeing in the gate.


mattt416:
Let's hold off on merging this for now, I have a funny feeling this isn't going to actually help debug the issue on hand.

mattt416:
This won't help us identify the specific issue mentioned, but we could still merge this for other reasons.  I'm happy to go either way.

PR #676, opened 2016-11-23
Remove jenkins-rpc AIO jobs

hughsaunders:
These jobs are irelevant as they test aio build script which is being
migrated away from. They also cause confusion as they do not test JJB
jobs :(

These jobs also rely on the non-JJB RPC-AIO job which is being removed.

A new strategy is required for testing jjb jobs, but it will involve
having a dev jenkins master, possibly even in docker.

Connects rcbops/u-suk-dev#700

### DIFF
--- a/rpc-jobs/jobs.yaml
+++ b/rpc-jobs/jobs.yaml
@@ -37,17 +37,6 @@
       - string:
           name: GIT_BRANCH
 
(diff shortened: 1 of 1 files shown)


hughsaunders:
recheck_jjb

hughsaunders:
recheck_jjb

hughsaunders:
I canceled the aio and upgrade jobs as this PR is to remove them..

PR #678, opened 2016-11-24
JJB Upgrade Matrix and Remove Jclouds Dep

hughsaunders:
The previous upgrade matrix job was a chain of Upgrade-Periodic ->
Upgrade Matrix --> RPC AIO.

This commit introduces JJB-Upgrade-Matrix which is a single job that
replaces all three previous jobs. The parallelisation of upgrades is
handled by a parallel block in a jenkins pipeline script rather than
seperate jobs. This new way of allocating nodes is not compatible with
jenkins jcloud single use slaves, as slaves are not reliably terminated
on completion. This would leave unclean slaves in the pool ready to
accept new jobs and that would cause future jobs to fail the node
cleanliness test.

Some major consequences of this approach:
  * The AIO and Upgrade Matrix Jobs used to call the same RPC-AIO job to
    do the majority of the work (but with different parameters). Now
    these jobs are both self contained, in order to prevent
    duplication, common code has been moved into bash_gating_lib.sh

  * As the jclouds approach to allocating nodes doesn't work well with
    pipeline script, the node creation/teardown is now handled by bash
    functions in gating_bash_lib. A positive of this is that its now
    possible to specify per job whether the node should be cleaned up.

Connects rcbops/u-suk-dev#353


hughsaunders:
Successful test PR job: http://jenkins.propter.net/job/JJB-RPC-AIO_master-cidev/89/
Successful test user initiated job: http://jenkins.propter.net/job/JJB-RPC-AIO_master-cidev/87

Let me know if any further information would assist with a review.

scripts/gating_bash_lib.sh
```diff
@@ -0,0 +1,612 @@
...
+  if [[ "${OA_REPO:-none}" != "none" ]]; then
+    pushd openstack-ansible
+      git remote add override $OA_REPO -f
+      git fetch -a
+      git checkout override/$OA_BRANCH
+    popd
+  fi
+}
+
+integrate_proposed_change(){
```
mattt416:
```
18:02:27 + integrate_proposed_change
18:02:27 + git rebase origin/do_not_build_on_pr
18:02:27 First, rewinding head to replay your work on top of it...
18:02:27 
18:02:27 *** Please tell me who you are.
18:02:27 
18:02:27 Run
18:02:27 
18:02:27   git config --global user.email "you@example.com"
18:02:27   git config --global user.name "Your Name"
18:02:27 
18:02:27 to set your account's default identity.
18:02:27 Omit --global to set the identity only in this repository.
18:02:27 
18:02:27 fatal: unable to auto-detect email address (got 'root@jramc-89-18028.(none)')
18:02:27 You need to set your committer info first
18:02:27 + git rebase --abort
18:02:27 No rebase in progress?
```

scripts/gating_bash_lib.sh
```diff
@@ -0,0 +1,612 @@
...
+    rackspace_cloud_password
+    rackspace_cloud_tenant_id
+    rackspace_cloud_username
+  )
+
+  #Dump env vars which will be required on the remote host
+  : > $WORKSPACE/ship.env
+  for env_key in ${ship_env_keys[@]}
+  do
+    echo "Storing $env_key=${!env_key:-}for shipping to instance"
```
mattt416: Needs a space here between the key and `for shipping`.

mattt416:
Hi @hughsaunders, I noticed that MaaS wasn't deployed in the two linked jobs, despite `DEPLOY_MAAS` being set to `yes`.  You can see later on when deploy.sh is run that `DEPLOY_MAAS=no`.  Do we need to export in ship.env or is there something else going on here?

mattt416:
This is a pretty heavy commit -- I'll try to look at it again once the mentioned issues are resolved.  One other thing I noticed looking at the jenkins console log was that we only build one horizon container now, where previously it was two.  Is that change deliberate?

mattt416:
Regarding the horizon count reducing, this was actually a change in stable/mitaka in openstack-ansible, so please disregard that comment :)

scripts/gating_bash_lib.sh
hughsaunders: I thought I'd fixed this with https://github.com/rcbops/jenkins-rpc/blob/bug/353/scripts/gating_bash_lib.sh#L127-L135 Further investigation replied.

hughsaunders: Added the git config commands into the git_prep function. Makes more sense to configure directly rather than messing with git's auto domain detection.

hughsaunders:
Running a non-maas test 😡 http://jenkins.propter.net/job/JJB-RPC-AIO_master-cidev/114/

mattt416:
Hi @hughsaunders, can you please address merge conficts and re-trigger a build so I can examine the console log output?

Thanks!

--Matt

hughsaunders:
Successful test run with current version (b63663f015b89f33863b66045b95c121359add68) http://jenkins.propter.net/job/JJB-RPC-AIO_mitaka-13.1-cidev/4/ Please review.

mattt416:
Let's hold off on merging this one until Hugh's back in the office tomorrow.

odyssey4me:
We can leave the merge action to @hughsaunders

scripts/gating_bash_lib.sh
```diff
@@ -0,0 +1,612 @@
...
+
+integrate_horizon_extensions_changes(){
+  # If horizon extensions are in tree and the appropriate PR trigger vars are
+  # set, override the horizon extensions git location so that the proposed
+  # changes are built.
+  # NOTE(hughsaunders): Any proposed changes to the horizon extensions won't be
+  # rebased or merged into master before being built. So if the proposed commit's
+  # parent is old, this test won't guarantee that it works with the current
+  # master.
+  if [[ -n "${ghprbAuthorRepoGitUrl:-}" && -n "${ghprbActualCommit:-}" ]]\
```
mattt416: Not sure how critical this is, but from http://jenkins.propter.net/job/JJB-RPC-AIO_mitaka-13.1-cidev/4/consoleFull:

```
...
11:03:29 + echo 'export ghprbAuthorRepoGitUrl=" "'
...
11:03:49 + echo 'export ghprbActualCommit=" "'
...
```

This means we're trying to git ls-tree against a space.

hughsaunders:
fire-in-the-hole-large-1485x667

PR #680, opened 2016-12-05
Ensure that no unwanted leftovers are in the artifact repo

odyssey4me:
The artifacts stored after a successful build may execute multiple
times and it is possible that it may leave some parts behind which
are unnecessary or not applicable.

For example:
The os-releases folder in a build may refer to a newer version of
a python package which happens to not be pinned by OpenStack's
upper constraints. In the artifact copy the older version of the
package will still be linked in the tag's os-releases folder, thus
referncing two possible versions fot that tag. This, in turn, will
create problems when executing a build using these artifacts
because the wrong python package version may be used.

This patch ensures that the os-releases and venv folders produced
in the build are copied to rpc-repo exactly, but also ensures that
the folders for other tags are left alone.

### DIFF
--- a/rpc-jobs/jobs.yaml
+++ b/rpc-jobs/jobs.yaml
@@ -429,6 +429,7 @@
                push(){{
                  src="$1"
                  dest="$2"
+                 extra_args=${{3:-}}
                  description="$src to rpc-repo:$dest"
                  echo "===== Start: $description  ===="
                  [[ -d $src ]] || {{
@@ -436,6 +437,7 @@
                    exit
                  }}
                  rsync \
+                   ${{extra_args}} \
                    --delay-updates \
                    --recursive \
                    --links \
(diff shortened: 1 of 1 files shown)


rpc-jobs/jobs.yaml
```diff
@@ -472,11 +472,17 @@
...
+               # The following artifacts are purely additive - nothing must
+               # ever be deleted at the target.
                push "${{local_base}}/pools/" "${{remote_base}}/pools"
-               push "${{local_base}}/os-releases/" "${{remote_base}}/os-releases"
                push "${{local_base}}/links/" "${{remote_base}}/links"
 
+               # The following artifacts must be exactly as they look in the
+               # build when it completes. Anything that is on the target that
+               # is not in the build must be removed.
+               push --delete "${{local_base}}/venvs/*" "${{remote_base}}/venvs/"
```
hughsaunders: "--delete" will be treated as the source by the push function, which I don't think is what you're after. 

The push function will need modification if you want to pass additional args to rsync.

rpc-jobs/jobs.yaml
```diff
@@ -429,13 +429,15 @@
                push(){{
                  src="$1"
                  dest="$2"
+                 extra_args=${3:-}
```
hughsaunders: "{}" is used for JJB var interpolation so must be escaped via doubling to be ignored by JJB and passed through to bash.

PR #681, opened 2016-12-05
Remove the cirros_tgz_url image for tempest

alextricity25:
This has been removed upstream for Newton[1], and tempest
hasn't been using them for some time.

[1] https://github.com/openstack/openstack-ansible/blob/stable/newton/releasenotes/notes/broader-image-support-69241983e5a36018.yaml#L22-L24

Connects https://github.com/rcbops/u-suk-dev/issues/618

### DIFF
--- a/scripts/aio_build_script.sh
+++ b/scripts/aio_build_script.sh
@@ -176,17 +176,18 @@ horizon_extensions_git_package_name: "horizon-extensions"
 EOVARS
 fi
 
-#Supply fixed cirros image while empty key bug is not fixed upstream.
+# Supply fixed cirros image while empty key bug is not fixed upstream.
+# Both tempest_img_url and cirros_img_url are overridden since
+# cirros_img_url was removed in the newton release, but is still
+# needed for our job to run tests againsts previous releases.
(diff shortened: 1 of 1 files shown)


alextricity25:
I'm also seeing strange tempest errors on the newton sha bump. tempest seems to be choosing the image file with the wrong version:
https://gist.github.com/alextricity25/87aa67f489bcbfa7470f8ee29dfde39f
http://jenkins.propter.net/job/JJB-RPC-AIO_master-ceph/1441/consoleFull

It's downloading 0.3.4, but it's trying to look for 0.3.1

mattt416:
Hi @alextricity25, I'm not sure this is correct because in stable/newton we're not overriding tempest_img_url and therefore tasks that use tempest_img_url are going to get the default image specified in defaults/main.yml.  Could you override tempest_img_url and cirros_img_url just so that this overrides fully irrespective of which branch is being deployed?

alextricity25:
Good catch @mattt416. I've updated the commit to include ``tempest_img_url`` so we can test both Newton and previous releases without the tempest playbooks failing.

PR #683, opened 2016-12-05
Remove upgrade job for master branch

alextricity25:
This is so we can start work on newton


hughsaunders:
This looks good to me, but required status for the upgrade job must be removed from rpc-openstack/master before this merges. 
screen shot 2016-12-06 at 18 44 00

alextricity25:
The upgrade job is no longer required on rpc-openstack. This PR is ready to merge.

PR #686, opened 2016-12-08
Add parameter to disable maas deploy

hughsaunders:
Useful for running test builds when MaaS is having problems.
Also adds user vars as a jenkins parameter so that maas_use_api can be
overriden.

Connects rcbops/u-suk-dev#810


alextricity25:
Does this turn off maas by default for all jobs? It looks like it does.

hughsaunders:
@alextricity25 no doesn't change the default, but does make it optional.

PR #692, opened 2016-12-14
Disable MaaS checks on Jenkins jobs for Newton (master) branch

prometheanfire:
MaaS checks are failing fairly often on the master branch of rpc-openstack.
This is slowing down development and tying up lots of Jenkins resources with
rechecks. We should disable MaaS checks in the master branch temporarily until
Newton is closer to its final release date.

*Commit message courtesy of @major*

### DIFF
--- a/rpc-jobs/jobs.yaml
+++ b/rpc-jobs/jobs.yaml
@@ -55,6 +55,7 @@
       - master:
           branch: master
           branches: "master"
+          USER_VARS: "maas_use_api: false"
     context:
       - swift
       - ceph:
@@ -84,6 +85,16 @@
           branches: "mitaka-13\\.[^0].*"
           UPGRADE: "yes"
           UPGRADE_FROM_REF: "origin/liberty-12.2"
+
+- project:
+    name: 'JJB-AIO-MAAS-Job'
+    jobs:
+      - 'JJB-RPC-AIO_{series}-{context}':
+          series: master
+          branch: master
+          branches: "do_not_build_on_pr"
+          context: periodic_maas
+          CRON: "H H * * *"
 
 - project:
     name: 'JJB-AIO-Test-Jobs'


alextricity25:
You can do that by passing in the ``USER_VARS`` instead of the ``DEPLOY_MAAS`` variable. Give it this value here:
```
USER_VARS: 	maas_use_api: false
```

major:
I'm totally in support of the disabling of MaaS temporarily in Newton until we're closer to release. If @alextricity25's method works, let's do that.

cloudnull:
+1 I think maas should be removed from the normal gate checks and moved into a periodic job.

BjoernT:
👍

rpc-jobs/jobs.yaml
```diff
@@ -86,6 +88,17 @@
           UPGRADE_FROM_REF: "origin/liberty-12.2"
 
 - project:
+  name: 'JJB-AIO-MAAS-Job'
+  jobs:
+    - 'JJB-RPC-AIO_{series}-{contex}':
+      series: master
+      branch: master
+      branches: "do_not_build_on_pr"
+      context: periodic
```
hughsaunders: context should be maas here, as this job is specifically for testing maas.

prometheanfire: From what I understand context is like a comment on the job.  So we'd want it to be descriptive.  If that's the case then I think it should be 'periodic_maas' (which I thought I set it to bug guess not).  If it's defining a category, you are right and it should probably just be 'maas' (though that's in the job name and the job name should probably be 'JJB-AIO-Periodic-MAAS-Job').

major: `periodic_maas` makes sense to me. It's a fairly good explanation of what the job is meant to do. ;)

major:
I think I've figured out the indentation shenanigans now.

rpc-jobs/jobs.yaml
```diff
@@ -55,6 +55,7 @@
       - master:
           branch: master
           branches: "master"
+          DEPLOY_MAAS: "no"
```
alextricity25: This needs to be replaced with ``USER_VARS: "maas_use_api: false"``. We are still deploy maas on master, we are just turning off the remote api checks.

rpc-jobs/jobs.yaml
```diff
@@ -86,6 +87,18 @@
...
+    name: 'JJB-AIO-MAAS-Job'
+    jobs:
+      - 'JJB-RPC-AIO_{series}-{context}':
+          series: master
+          branch: master
+          branches: "do_not_build_on_pr"
+          context: periodic_maas
+          CRON: "H H * * *"
+          DEPLOY_MAAS: "yes"
+          USER_VARS: "maas_use_api: false"
```
alextricity25: This line needs to be removed. The periodic job *will* test remote api checks.

alextricity25:
this-is-fine-meme

PR #699, opened 2017-01-03
Move heat stack deletion to post build action

hughsaunders:
This will allow cleanup to happen even when a build times out.
Also increases the timeout.

Connects rcbops/u-suk-dev#918


hughsaunders:
This needs to be merged with https://github.com/rcbops/rpc-heat-ansible/pull/6

PR #702, opened 2017-01-04
Introducing apt artifact building type of jobs

evrardjp:
This is a commit to introduce apt frozen repo building for
RPC release.

### DIFF
--- a/rpc-jobs/jobs.yaml
+++ b/rpc-jobs/jobs.yaml
@@ -632,6 +632,89 @@
       - junit:
           results: "artifacts/*.xml"
           allow-empty-results: true
+
+- job:
+    name: JJB-apt-artifacts
+    node: master
+    concurrent: true
+    build-discarder:
+      daysToKeep: 30
+    triggers:
+      - timed: "H 4 * * *"
+    parameters:
+    # - pr-params
+    # - string:
+    #     name: ghprbTargetBranch
+    #     default: "{branch}"
+    #     description: |
+    #       Target branch - the branch to be tested (sha1 param) will
+    #       be rebased against this. Also overridden by github pull request builder plugin.
+      - string:
+          name: RPC_ARTIFACTS
+          default: "https://github.com/rcbops/rpc-artifacts.git"
+          description: "RPC-Artifacts Repo"
+      - string:
+          name: RPC_ARTIFACTS_BRANCH
+          default: "master"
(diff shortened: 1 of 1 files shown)


rpc-jobs/jobs.yaml
```diff
@@ -632,6 +632,90 @@
...
+      - git:
+          url: "${RPC_ARTIFACTS}"
+          branches:
+            - "${RPC_ARTIFACTS_BRANCH}"
+    builders:
+      - shell: |
+          #!/bin/bash
+          #git clone ${RPC_ARTIFACTS} rpc-artifacts
+          #cd rpc-artifacts
+          export ANSIBLE_DEBUG=${ANSIBLE_DEBUG}
```
hughsaunders: ANSIBLE_DEBUG should also be an environment variable set by Jenkins, so I don't think this is doing anything.

rpc-jobs/jobs.yaml
```diff
@@ -632,6 +632,90 @@
...
+          mkdir -p ~/.ssh/
+          cat $RPC_REPO_KEY > ~/.ssh/repo.key
+          cat $REPO_GPG_PRIVATE > /openstack/aptly.private.key
+          cat $REPO_GPG_PUBLIC > /openstack/aptly.public.key
+          set -x
+          apt-get update
+          xargs apt-get install -y < bindep.txt
+          curl https://bootstrap.pypa.io/get-pip.py | python
+          pip install ansible==2.2
+          echo "[mirrors]" >> inventory
```
hughsaunders: Is there anything already in the inventory? Otherwise > should be used after the first echo to ensure the file is overwritten.

rpc-jobs/jobs.yaml
```diff
@@ -632,6 +632,90 @@
       - junit:
           results: "artifacts/*.xml"
           allow-empty-results: true
+
+- job:
+    name: JJB-apt-artifacts
+    node: master
```
hughsaunders: This is fine for now, but will have to change when we migrate.

rpc-jobs/jobs.yaml
evrardjp: Ahah fair point.

rpc-jobs/jobs.yaml
evrardjp: There is. We are fetching a simple inventory, we are completing it with what we need.

PR #703, opened 2017-01-04
[DONOTMERGE] Test MaaS on all periodics

mattt416:
Currently, MaaS is *only* being tested on the master-specifc MaaS
periodic job. MaaS is not being tested on kilo, liberty, or mitaka nor
on any upgrade jobs.

If the general consensus is that MaaS is too problematic to test on
per-commit builds, we should ensure that MaaS is tested on the periodic
jobs that run every 6 hours.  This means we can remove the
MaaS-specific periodic and ensure that we have MaaS being tested on
each series.

Please note that this change may not be the best way to accomplish
this, however this was the only way I can see how to implement this
without duplicating the job-template.

Connects https://github.com/rcbops/u-suk-dev/issues/867

**PLEASE DO NOT MERGE THIS YET AS IT'S LIKELY TO BREAK JENKINS**


mattt416:
**PLEASE DON'T MERGE THIS YET**

PR #704, opened 2017-01-05
Merge upgrade project and remove MaaS periodic

mattt416:
This refactor does the following:

1. Merges JJB-AIO-Upgrades into JJB-AIO-Jobs project and sets necessary
   excludes so we only get upgrades on the appropriate branches
2. Adds a new ztrigger axis for periodic and pr triggers which allows
   us to set DEPLOY_MAAS for a given trigger type
3. Renames the job template to JJB-RPC-AIO_{series}-{context}-{ztrigger}
4. Sets the default for DEPLOY_MAAS="yes" in the job template and then
   sets DEPLOY_MAAS="no" for the pr trigger in the project
5. Removes the JJB-AIO-MAAS-Job project since this is no longer needed

Rationale:

Currently, MaaS is *only* being tested on the master-specifc MaaS
periodic job. MaaS is not being tested on kilo, liberty, or mitaka nor
on any upgrade jobs. This refactor is being done primarily to ensure
that MaaS is being tested on all periodic jobs while continuing to be
skipped on all pr jobs.

Connects https://github.com/rcbops/u-suk-dev/issues/867

### DIFF
--- a/rpc-jobs/jobs.yaml
+++ b/rpc-jobs/jobs.yaml
@@ -57,6 +57,7 @@
       - mitaka:
           branch: mitaka-13.1
           branches: "mitaka-.*"
+          UPGRADE_FROM_REF: "origin/liberty-12.2"
       - master:
           branch: master
           branches: "master"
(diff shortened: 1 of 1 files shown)


mattt416:
This is not ready to merge, the excludes don't seem to be working as I'm not seeing *any* kilo or master periodic/pr jobs.

PR #705, opened 2017-01-09
Try secrets files

evrardjp:


rpc-jobs/jobs.yaml
```diff
@@ -670,10 +670,10 @@
         - text:
             credential-id: RPC_REPO_KEY
             variable: REPO_KEY
-        - text:
+        - secret-file:
```
hughsaunders: should be -file rather than secret-file

rpc-jobs/jobs.yaml
```diff
@@ -670,10 +670,10 @@
...
             credential-id: REPO_GPG_SECRET_FILE
             variable: REPO_GPG_PRIVATE
-        - text:
+        - Secret file:
```
hughsaunders: should be -file rather than secret-file

hughsaunders:
example here is comprehensive http://docs.openstack.org/infra/jenkins-job-builder/wrappers.html#wrappers.credentials-binding

evrardjp:
Done.

PR #707, opened 2017-01-10
Fix scm for apt artifacts

evrardjp:
The apt artifacts used a variable in scm but that variable was
commented out in a previous commit.

This should fix the regression.

### DIFF
--- a/rpc-jobs/jobs.yaml
+++ b/rpc-jobs/jobs.yaml
@@ -632,21 +632,21 @@
     triggers:
       - timed: "H 4 * * *"
     parameters:
-    # - pr-params
-    # - string:
-    #     name: ghprbTargetBranch
-    #     default: "{branch}"
-    #     description: |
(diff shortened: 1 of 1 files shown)


rpc-jobs/jobs.yaml
```diff
@@ -632,21 +632,21 @@
...
-    #     name: RPC_ARTIFACTS
-    #     default: "https://github.com/rcbops/rpc-artifacts.git"
-    #     description: "RPC-Artifacts Repo"
-    # - string:
-    #     name: RPC_ARTIFACTS_BRANCH
-    #     default: "master"
-    #     description: "RPC-Artifacts Branch"
+      - pr-params
+      - string:
+          name: ghprbTargetBranch
```
hughsaunders: is this used?

hughsaunders: merged anyway, we can remove it if its not required.

PR #708, opened 2017-01-10
Reuse existing secret for SSH key

evrardjp:
SSH Key file is already a stored secret, so we should reuse it in
apt artifacts, instead of having another one.


evrardjp:
What do you think of this?

It could be better for security to separate the keys, because different usage, but it's also more secrets to manage.

What's your opinion?

PR #712, opened 2017-01-12
Update AIO Cleanup script

hughsaunders:
The AIO cleanup script now reflects our curent AIO deployment naming and
methodology. Logic related to jenkins slave status has been removed
because job instances are no longer registerd as slaves.

The old threshold has been increased to 48 hours so that if a job is run
with keep_instance then the requester has two days to use the instance
for deugging before its removed.

Also:
 * moves AIO template defaults into a global defaults object
 * Adds parameters to the jenkins-rpc-git macro

Connects rcbops/u-suk-dev#953

### DIFF
--- a/rpc-jobs/jobs.yaml
+++ b/rpc-jobs/jobs.yaml
@@ -1,3 +1,34 @@
+
+# defaults - These must be set as the default value of a jenkins parameter in order to be injected into the build Environment.
+- defaults:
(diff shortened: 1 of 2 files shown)


hughsaunders:
successful test: http://jenkins.propter.net/job/JJB-Misc-AIO-Cleanup/17790/

PR #714, opened 2017-01-12
Fail on first error in apt artifacts

evrardjp:
If error happens, the build should fail immediately. This wasn't
done for apt artifacts.


evrardjp:
Do not merge yet -- testing.

evrardjp:
This is ready!

PR #715, opened 2017-01-12
Use artifacts folder variable

evrardjp:
This commit moves to use a jenkins variable to define the artifact
folder. The environment variables are available everywhere (bash,
ansible) and therefore it's easier to use it as single source of
truth: if the folder location changes, we have one variable to
change in jenkins, instead of fixing the path in different repos
(like rpc-artifacts, jenkins-rpc).

### DIFF
--- a/rpc-jobs/jobs.yaml
+++ b/rpc-jobs/jobs.yaml
@@ -655,6 +655,10 @@
           name: ANSIBLE_VERBOSITY
           default: "-v"
           description: "Change the verbosity level of ansible playbook runs"
+      - string:
+          name: RPC_ARTIFACTS_FOLDER
+          default: "/var/www/artifacts"
+          description: "Folder to store artifacts (private and public)"
     wrappers:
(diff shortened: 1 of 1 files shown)


rpc-jobs/jobs.yaml
```diff
@@ -655,6 +655,10 @@
           name: ANSIBLE_VERBOSITY
           default: "-v"
           description: "Change the verbosity level of ansible playbook runs"
+      - string:
+          name: RPC_ARTIFACTS_FOLDER
+          default: "/var/www/artifacts"
```
mattt416: The value is changing, is that expected?  Also, do we want to keep private keys under /var/www?

hughsaunders: The change is required because of the disk layout on repo1. Private keys under /var/www is ok because the doc root is /var/www/repo.

evrardjp:
Agreed with what you both said.

PR #716, opened 2017-01-12
Fix list of builded rpc artifacts

evrardjp:
The builded RPC artifacts are not residing in /openstack anymore,
but follows a variable instead.


evrardjp:
recheck

evrardjp:
recheck_all

PR #717, opened 2017-01-12
Add global defaults to JJB-Job-Test

mattt416:
This commit adds the global defaults to the JJB-Job-Test job so that it
can properly resolve JENKINS_RPC_REPO and JENKINS_RPC_BRANCH.

### DIFF
--- a/rpc-jobs/jobs.yaml
+++ b/rpc-jobs/jobs.yaml
@@ -518,6 +518,7 @@
 
 - job:
     name: JJB-Job-Test
+    defaults: global
     project-type: freestyle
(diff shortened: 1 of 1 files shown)


mattt416:
@hughsaunders This check is going to fail because the deployed job in Jenkins is incorrect.  I'm not sure if that needs to be corrected by hand so that this can pass?

Side note, there are other jobs in jobs.yaml that don't have the global defaults set.  I'm not sure right now if they need it or not.

evrardjp:
This is not needed, we merged the templating in the job job-test because it's not a template.

PR #718, opened 2017-01-13
Remove template vars from non-template jobs

hughsaunders:
Template vars are only processed in job templates and projects, not in
standalone jobs.

Connects rcbops/u-suk-dev#975


hughsaunders:
recheck_all

evrardjp:
👍

PR #724, opened 2017-01-16
Fix SCM references in JJB test/update jobs

hughsaunders:
These were broken in an attempt to use a macro for scm configuration, however that macro was for jobs that use jenkins-rpc as a secondary repo, not the one which is checked out with the changes to be tested. 

Also 
  * fixes a "yes" that should have been quoted to avoid autoboolification
  * Adds a git show HEAD so that the correct commit can be verified. The commit that's checkout out is the merge commit from the PR branch, which doesn't match the actual commit at the tip of the PR branch.                                                                                                                                                                                                  

Connects rcbos/u-suk-dev#981

### DIFF
--- a/rpc-jobs/jobs.yaml
+++ b/rpc-jobs/jobs.yaml
@@ -516,23 +516,36 @@
           default: "https://github.com/rcbops/jenkins-rpc"
           description: "Repo url for JENKINS_RPC"
       - string:
-          name: JENKINS_RPC_BRANCH
+          name: sha1
           default: "master"
-          description: "Branch to checkout for JENKINS_RPC"
+          description: |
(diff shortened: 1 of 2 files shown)


hughsaunders:
recheck_all

hughsaunders:
recheck_all

hughsaunders:
recheck_all

hughsaunders:
recheck_all

hughsaunders:
Hold off on this for a minute and I'll add in the credentials binding so that the log is starred out.

hughsaunders:
Now ready for merge, succesfull test (see above) and [update](http://jenkins.propter.net/job/JJB-Job-Update/41/console) jobs have run.

PR #727, opened 2017-01-18
Add periodic job to update rpc-repo git repositories

odyssey4me:
This patch adds a periodic job to update the git repositories
hosted in rpc-repo based on all the sources used in the master
of openstack-ansible.

This is to ensure that regardless of the series being
implemented, rpc-repo should have a copy of all the possible
git repositories required.

If, at a later stage, it is found that the repo server needs
to have additional repositories, then the job/playbooks will
be adjusted to accommodate this.

### DIFF
--- a/rpc-jobs/jobs.yaml
+++ b/rpc-jobs/jobs.yaml
@@ -789,6 +789,76 @@
           ls -R ${RPC_ARTIFACTS_PUBLIC_FOLDER}
 
 - job:
+    name: JJB-artifacts-git
+    node: ArtifactBuilder
+    concurrent: false
+    build-discarder:
+      daysToKeep: 30
+    triggers:
+      - timed: "H 4 * * *"
+    parameters:
+      - string:
+          name: GIT_RPC_ARTIFACTS
+          default: "https://github.com/rcbops/rpc-artifacts"
+          description: "RPC-Artifacts Git Repository"
+      - string:
+          name: GIT_OSA
+          default: "https://github.com/openstack/openstack-ansible"
+          description: "OpenStack-Ansible Git Repository"
+    wrappers:
+      - ansicolor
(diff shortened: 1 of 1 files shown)


rpc-jobs/jobs.yaml
```diff
@@ -789,6 +789,68 @@
           ls -R ${RPC_ARTIFACTS_PUBLIC_FOLDER}
 
 - job:
+    name: JJB-artifacts-git
+    node: master
```
hughsaunders: Could you run this on node ArtifactBuilder ? It has more resources for dealing with artifacts.

rpc-jobs/jobs.yaml
```diff
@@ -789,6 +789,68 @@
...
+          mkdir -p ~/.ssh/
+          set +x
+          cat $REPO_USER_KEY > ~/.ssh/repo.key
+          chmod 600 ~/.ssh/repo.key
+          set -x
+          grep "${REPO_HOST}" ~/.ssh/known_hosts || echo "${REPO_HOST} $(cat $REPO_HOST_PUBKEY)" >> ~/.ssh/known_hosts
+          # We need both Ansible and the OSA plugins available.
+          # The simplest way to do this is to use the bootstrap script.
+          ./scripts/bootstrap-ansible.sh
+          git clone ${GIT_RPC_ARTIFACTS} /opt/rpc-artifacts/
```
hughsaunders: This may fail if the workspace is not clean (which it won't be on the second run).

odyssey4me: @evrardjp FYI there will never be a need to use an alternative branch for OSA. The intent here is to always only grab the latest list of repositories from the master branch. I cannot fathom any reason why you'd want to target an alternative branch.

PR #731, opened 2017-01-20
Ensure that role downloads happen via git

odyssey4me:
The aptly job downloads the roles via git. To prevent a
race condition with that job we ensure that the roles are
downloaded via git.


hughsaunders:
recheck_all

PR #732, opened 2017-01-20
Re-enable MaaS tests in Jenkins

major:
Now that lots of work is underway in rcbops/u-suk-dev#948 to
improve MaaS, we should re-enable the tests in Jenkins.

### DIFF
--- a/rpc-jobs/jobs.yaml
+++ b/rpc-jobs/jobs.yaml
@@ -117,7 +117,7 @@
     ztrigger:
       - pr:
           CRON: ""
-          DEPLOY_MAAS: "no"
+          USER_VARS: "maas_use_api: false"
       - periodic:
           branches: "do_not_build_on_pr"
     exclude:


rpc-jobs/jobs.yaml
```diff
@@ -117,7 +117,7 @@
     ztrigger:
       - pr:
           CRON: ""
-          DEPLOY_MAAS: "no"
+          DEPLOY_MAAS: "yes"
```
hughsaunders: This is default so you can remove the line. 

If you want to enable MaaS for a specific job type  (eg swift) you could set DEPLOY_MAAS: "no" under every context apart from swift.

mattt416:
I say we try setting `maas_use_api: false` and removing the `DEPLOY_MAAS: no`.

major:
Where does `use_maas_api` belong, @mattt416?

rpc-jobs/jobs.yaml
mattt416: You can add the following here I believe:

USER_VARS: "maas_use_api: false"

major: Okay, I'll give that a try.

PR #733, opened 2017-01-20
Change periodic trigger time

mattt416:
Currently, all periodics are triggered every 6 hours.  This commit
updates the default schedule to trigger jobs once between 9 AM and
9 PM UTC (giving us reasonable coverage across UK and US shifts),
and only runs them on week days.

The intent here is to run the periodics less frequently, allowing us to
pay more attention to failures and be quicker to respond to issues.


hughsaunders:
@mattt416 Jenkins.propter is on UTC

PR #734, opened 2017-01-25
Allow the destruction of apt artifacts

evrardjp:
Currently if the job re-runs, the apt artifacts for the same build
number will not be created again (on purpose). We want to have
the feature to override this in the CI. The code in rpc-artifacts
was done, but no wiring existed in jenkins. This should fix it.


mattt416:
LGTM

PR #735, opened 2017-01-27
Add python artifact generation job

odyssey4me:
This patch adds some moving parts to the scripts library
and adjusts the AIO job template to be able to use it
for the building of python artifacts.

A future additional job will use a similar mechanism to
build container artifacts.

Connects https://github.com/rcbops/u-suk-dev/issues/1020

### DIFF
--- a/rpc-jobs/jobs.yaml
+++ b/rpc-jobs/jobs.yaml
@@ -28,6 +28,7 @@
     CRON: "H H(9-21) * * 1-5"
     KEEP_INSTANCE: "no"
     IMAGE: "Ubuntu 14.04 LTS"
+    BUILD_SCRIPT_FUNCTION: "managed_aio"
 
 #
 # Macros for repeated blocks
@@ -179,6 +180,22 @@
             set -e
             bash scripts/heat_cleanup.sh
 
+- project:
+    name: 'JJB-artifacts-python'
+    series:
+      - master:
+          branch: master
+          branches: master
+      - newton:
+          branch: newton-14.0
+          branches: "newton-.*"
+    jobs:
+      - 'JJB-RPC-AIO_{series}-{context}-{ztrigger}':
+          context: experimental
+          ztrigger: ondemand
+          DEPLOY_MAAS: "no"
+          CRON: ""
+          BUILD_SCRIPT_FUNCTION: managed_aio_artifacts
 
 ## Job Definitions
 # This template is for testing PRs against Jenkins-RPC
@@ -302,6 +319,10 @@
           default: "{BUILD_SCRIPT_PATH}"
           description: "Path to the build script within JENKINS_RPC_REPO"
       - string:
+          name: BUILD_SCRIPT_FUNCTION
+          default: "{BUILD_SCRIPT_FUNCTION}"
+          description: "The name of the build function within JENKINS_RPC_REPO"
+      - string:
           name: OA_REPO
           default: "{OA_REPO}"
           description: "Openstack Ansible Repo - can be used to override the RPC OSA submodule"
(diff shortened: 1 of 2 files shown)


rpc-jobs/jobs.yaml
```diff
@@ -354,7 +358,7 @@
             git checkout $JENKINS_RPC_BRANCH
             git reset --hard origin/$JENKINS_RPC_BRANCH
             . scripts/gating_bash_lib.sh
-            managed_aio "{IMAGE}"
+            {BUILD_SCRIPT_NAME} "{IMAGE}"
```
hughsaunders: build_function? its not really a script path.

hughsaunders: both of these should be $foo or ${{ foo }}. IMAGE happens to work as its defined as an axis in a project, however in its current state it won't be overrideen if the user specifies a different image when initiating a build.

odyssey4me: @hughsaunders I need help building the actual job that uses this template. We can start with a periodic job that executes against the master branch only, but in the long term we'll want this job to execute based on the following triggers:

1. When a new branch is created of the form ``<series name>-<RPC major.minor version tag>``
2. When a tag is published

The intent of the job is to publish the python artifacts for the next tag that will be released.

scripts/gating_bash_lib.sh
```diff
@@ -103,6 +103,48 @@ aio(){
...
+}
+
+aio_artifact_build(){
+  prep
+  sudo -E\
+    TERM=linux \
+    DEPLOY_AIO=yes \
+    scripts/bootstrap-ansible.sh \
+    scripts/bootstrap-aio.sh \
+    cd /opt/rpc-openstack/openstack-ansible/playbooks \
```
odyssey4me: Need to add a step here to execute the rpc-artifacts script that guesses the next release tag.

scripts/gating_bash_lib.sh
```diff
@@ -103,6 +103,48 @@ aio(){
...
+  return $deploy_result
+  # teardown called by exit handler
+}
+
+aio_artifact_build(){
+  prep
+  sudo -E\
+    TERM=linux \
+    DEPLOY_AIO=yes \
+    scripts/bootstrap-ansible.sh \
```
odyssey4me: I expect that we need a ``;`` after each command, and perhaps all the commands require wrapping in ``"`` to ensure that they're all passed to sudo?

odyssey4me:
@hughsaunders I need help building the actual job that uses this template. We can start with a periodic job that executes against the master branch only, but in the long term we'll want this job to execute based on the following triggers:

1. When a new branch is created of the form ``<series name>-<RPC major.minor version tag>``
2. When a tag is published

The intent of the job is to publish the python artifacts for the next tag that will be released.

rpc-jobs/jobs.yaml
```diff
@@ -302,6 +302,10 @@
           default: "{BUILD_SCRIPT_PATH}"
           description: "Path to the build script within JENKINS_RPC_REPO"
       - string:
+          name: BUILD_SCRIPT_NAME
+          default: "managed_aio"
```
melissa-kam: You'll want to change this to "{BUILD_SCRIPT_NAME}" if you want to create both managed_aio and managed_aio_artifacts jobs from this same template.

scripts/gating_bash_lib.sh
hughsaunders: As you're using sudo -E, you can export the env vars before sudo, then sudo each script individually.

evrardjp:
@hughsaunders what's your opinion on ``context: experimental`` and ``ztrigger: ondemand``? 
I don't need to override things with ztrigger, and basically context could be anything for me if I'm not mistaken.

scripts/gating_bash_lib.sh
```diff
@@ -103,6 +103,43 @@ aio(){
...
+  on_remote aio_artifact_build
+  deploy_result=$?
+  # work out working dir on remote, for executing stuff and collecting artefacts
+  return $deploy_result
+  # teardown called by exit handler
+}
+
+aio_artifact_build(){
+  prep
+  TERM=linux
```
hughsaunders: this won't be passed through sudo -E unless its exported, eg `export TERM=linux`

PR #742, opened 2017-02-01
Switch git update job to use RPC-O

odyssey4me:
RPC-O requires more git repositories than OSA, so this
patch switches the git repo update job to ensure that
RPC's list of git repositories is used instead of just
OSA.

Connects https://github.com/rcbops/u-suk-dev/issues/1111

### DIFF
--- a/rpc-jobs/jobs.yaml
+++ b/rpc-jobs/jobs.yaml
@@ -817,9 +817,13 @@
           default: "https://github.com/rcbops/rpc-artifacts"
           description: "RPC-Artifacts Git Repository"
       - string:
-          name: GIT_OSA
-          default: "https://github.com/openstack/openstack-ansible"
-          description: "OpenStack-Ansible Git Repository"
+          name: RPC_REPO
+          default: "{RPC_REPO}"
+          description: "RPC REPO URL"
+      - string:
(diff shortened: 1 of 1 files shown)


rpc-jobs/jobs.yaml
```diff
@@ -859,6 +863,10 @@
           # We want the role downloads to be done via git
           # This ensures that there is no race condition with the aptly job
           export ANSIBLE_ROLE_FETCH_MODE="git-clone"
+          # The rpc-openstack repo requires the openstack-ansible submodule
+          # to be initiated and updated
+          git submodule init
+          git submodule update
           # We need both Ansible and the OSA plugins available.
           # The simplest way to do this is to use the bootstrap script.
           ./scripts/bootstrap-ansible.sh
```
mattt416: Does some of this stuff assume you're in openstack-ansible and not rpc-openstack?

odyssey4me: @mattt416 Nope - ``./scripts/bootstrap-{aio,ansible}.sh`` are available in the ``master`` and ``newton-14.0`` branches, which are the only ones I care about for this job.

odyssey4me:
recheck_all

PR #744, opened 2017-02-01
Set the BASE_DIR for the git artifacts job

odyssey4me:
Without setting this, it is assumed to be '/opt/rpc-openstack'
which is not where this checkout is located.

Connects https://github.com/rcbops/u-suk-dev/issues/1111


odyssey4me:
recheck_all

odyssey4me:
recheck_all

PR #746, opened 2017-02-06
Use rpc-openstack artifact scripts for git updates

odyssey4me:
Previously the rpc-artifacts repository was used.
This patch switches the job to use the scripts in
the rpc-openstack repository.

Connects https://github.com/rcbops/u-suk-dev/issues/1131

### DIFF
--- a/rpc-jobs/jobs.yaml
+++ b/rpc-jobs/jobs.yaml
@@ -813,16 +813,12 @@
       - timed: "H 4 * * *"
     parameters:
       - string:
-          name: GIT_RPC_ARTIFACTS
-          default: "https://github.com/rcbops/rpc-artifacts"
-          description: "RPC-Artifacts Git Repository"
-      - string:
           name: RPC_REPO
           default: "https://github.com/rcbops/rpc-openstack"
           description: "RPC REPO URL"
       - string:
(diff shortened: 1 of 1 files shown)


rpc-jobs/jobs.yaml
```diff
@@ -873,20 +869,12 @@
...
-              git checkout master
-              git reset --hard origin/master
-              git pull
-            popd
-          fi
-          # Append rpc-repo host to [mirrors] group
-          echo "repo ansible_host=${REPO_HOST} ansible_user=${REPO_USER} ansible_ssh_private_key_file='~/.ssh/repo.key' " >> /opt/rpc-artifacts/inventory
+          # Setup the inventory
+          echo '[mirrors]' > ${BASE_DIR}/inventory
+          echo "repo ansible_host=${REPO_HOST} ansible_user=${REPO_USER} ansible_ssh_private_key_file='~/.ssh/repo.key' " >> ${BASE_DIR}/inventory
```
mattt416: Wonder if it may make sense to dump the inventory into ${BASE_DIR}/scripts/artifacts-building, just so that it's out of the way and won't cause any issues elsewhere.

PR #748, opened 2017-02-07
Switch to using rpc-o for apt artifact job

odyssey4me:
Previously the apt artifact build job used the rpc-artifacts
repository. This patch switches it to use the rpc-openstack
repository instead.

Connects https://github.com/rcbops/u-suk-dev/issues/1143

### DIFF
--- a/rpc-jobs/jobs.yaml
+++ b/rpc-jobs/jobs.yaml
@@ -737,13 +737,13 @@
             Target branch - the branch to be tested (sha1 param) will
             be rebased against this. Also overridden by github pull request builder plugin.
       - string:
-          name: RPC_ARTIFACTS
-          default: "https://github.com/rcbops/rpc-artifacts.git"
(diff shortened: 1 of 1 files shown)


rpc-jobs/jobs.yaml
```diff
@@ -791,12 +791,12 @@
...
       - git:
-          url: "${RPC_ARTIFACTS}"
+          url: "${RPC_REPO}"
           branches:
-            - "${RPC_ARTIFACTS_BRANCH}"
+            - "${RPC_REPO_BRANCH}"
     builders:
       - shell: |
-          ./build-apt-artifacts.sh
+          ./scripts/artifacts-building/apt/build-apt-artifacts.sh
```
evrardjp: We need to make sure this is cloned on the right folder, because I'm refering to ``/opt/rpc-openstack`` in the ``build-apt-artifacts.sh`` script.

PR #750, opened 2017-02-13
[WIP] Test lvm and ceph backends when ceph deployed

mattt416:
This commit makes some tempest-related overrides such that tempest will
test both ceph and lvm backends when ceph is deployed.  This necessary
as we recently had an issue with cinder-volume not being able to speak
to the ceph cluster which went largely undetected.


mattt416:
Testing this in http://jenkins.propter.net/job/JJB-RPC-AIO_newton-ceph-periodic/60/console

mattt416:
This doesn't work atm, created upstream [review](https://review.openstack.org/#/c/434536/1) to address.

PR #752, opened 2017-03-09
Disable master jobs

hughsaunders:
The first jobs to transition to CIT jenkins are the master jobs,
this commit removes them from jenkins-rpc.

Its important this is not merged before rcbops/rpc-openstack#1991

Connects rcbops/u-suk-dev#1313


hughsaunders:
we'll do this whole sale rather than branch by branch

PR #756, opened 2017-03-30
Remove git/apt/python/container jobs

odyssey4me:
Now that the python/container artifact builds are
in rpc-gating we don't need them here.

The apt and git jobs are also moving over shortly
and we don't need them to run on the old system
until they're up on the new system.

Connects https://github.com/rcbops/u-suk-dev/issues/1413
Connects https://github.com/rcbops/u-suk-dev/issues/1451
Connects https://github.com/rcbops/u-suk-dev/issues/1454


evrardjp:
Ok let's remove their job definition. We'll still have to leave the https://github.com/rcbops/u-suk-dev/issues/1454 (and an equivalent for artifact node 2) open for now.

